PR #51, opened 2026-04-16
feat: Add --dry-run for manual image generation & migrate scripts to uv

kanazawahere:
This PR introduces a \--dry-run\ flag to \image_gen.py\. It exports the prompt payloads into an \image_prompts.json\ file instead of making API calls, which allows users to manually generate images externally (like via Midjourney or Flux) and then drop them into the project. This PR also comprehensively migrates all markdown documentation and python wrapper scripts to formally adopt \uv\ for fast and seamless dependency management.


hugohe3:
Thanks for the contribution — appreciate the effort on both fronts.

**On `--dry-run`:** I like the idea. The "export prompts → generate externally via Midjourney/Flux/Gemini web → drop images back" workflow fits how this project is actually used, and it pairs well with the existing `image_prompts.md` output from the Image_Generator role. One blocker before merge: `_resolve_backend()` is called before the dry-run branch, so users still need `IMAGE_BACKEND` set and the backend SDK installed just to export prompts — which defeats the main use case. If you move `_resolve_backend()` inside the `if not args.dry_run` path (and let `backend`/`model` fall back to `None` in the JSON), this part is good to go.

**On the uv migration:** I'm going to pass on this part, and I want to be transparent about why so it doesn't feel arbitrary.

The target audience for this project is domain experts (consultants, analysts, researchers) across multiple AI IDEs — not Python tooling power users. The README deliberately promises "you only need Python" as the single install step. Promoting `uv` to a required dependency adds a second tool to learn, a second install path (especially painful on Windows), and interacts awkwardly with Cursor / VS Code / Codebuddy's built-in venv+pip integration. For a repo with this many forks, that's a meaningful onboarding tax.

`uv` is a great tool, and users who already know it can run `uv pip install -r requirements.txt` against the existing `requirements.txt` without any repo-level changes — they don't need official endorsement to benefit. Conversely, making it the default path would break the "minimum viable setup" promise for everyone else.

A few concrete issues in the current migration also reinforce this: `pyproject.toml` pins `requires-python = \">=3.12\"` while the README still promises 3.10+ (silent breakage for 3.10/3.11 users); `cairosvg` (the recommended Option A for SVG→PNG fallback) is missing from `pyproject.toml`; `requirements.txt` and `pyproject.toml` both ship and will drift; and `update_repo.py` triggers on `requirements.txt` changes but now runs `uv sync` against `pyproject.toml`. These are all fixable, but they're symptoms of the larger positioning question, and fixing them still leaves the UX tradeoff unchanged.

**Proposed path forward:** could you split this into two PRs?

1. \`--dry-run\` only, with the backend-resolution fix — happy to merge.
2. The uv migration — I'll close with this explanation, no reflection on the work.

Thanks again for engaging with the project.

PR #52, opened 2026-04-17
feat: switch MiniMax image backend to international endpoint and add tests

ximiximi423:
## Summary

The MiniMax image generation backend currently defaults to `api.minimaxi.com`, the domestic (China) endpoint. For global users outside China, this can cause connectivity issues. This PR switches the default to the international endpoint `api.minimax.io` and makes several improvements:

- **Default endpoint**: `api.minimaxi.com` → `api.minimax.io` (international/overseas)  
  Domestic users can still use the old endpoint by setting `MINIMAX_BASE_URL=https://api.minimaxi.com`
- **Bug fix in `_resolve_url()`**: the previous implementation returned a double `/v1/v1/image_generation` path if `MINIMAX_BASE_URL` was set to a versioned base (e.g. `https://api.minimax.io/v1`). Now handles all three URL forms correctly:
  - full endpoint (`…/image_generation`) → used as-is
  - versioned base (`…/v1`) → appends `/image_generation`
  - root base (`https://api.minimax.io`) → appends `/v1/image_generation`
- **Unit tests** (`test_backend_minimax.py`, 19 assertions): URL resolution, dimension mapping, and image-bytes extraction
- **Documentation** (`docs/image.md`): add MiniMax `.env` example with a note about the domestic override
- **Model name correction**: README and README_CN.md updated from `MiniMax 2.7` to the proper API identifier `MiniMax-M2.7`

The international endpoint was verified with a live image generation call (`image-01` model, 512×512, 12.6 s response time).

## API references

- MiniMax image generation: https://platform.minimax.io/docs/api-reference/image-generation
- MiniMax chat (for reference): https://platform.minimax.io/docs/api-reference/text-openai-api


hugohe3:
First off — really high-quality contribution here 👏

The three-form URL handling in `_resolve_url` is thoughtfully done, the docstring lays out each case clearly, and the tests are clean, well-named, and precisely scoped. The PR description covers motivation, changes, and verification thoroughly. It's clear care went into this, and I appreciate it 🙏

A few things I'd like to discuss, mostly about staying consistent with the rest of the repo:

**1. On the test file — I'd like to ask you to remove it**

The repo currently has **zero test infrastructure** — no other test files anywhere, no CI, no `requirements-dev.txt`, no docs explaining how to run tests. This isn't an oversight; it's a deliberate posture for the project: solo-maintained, limited bandwidth, no plan to invest in test infra right now.

If we merge this test file in, it becomes the **only** test in the repo — nothing will run it automatically, and future contributors editing `backend_minimax.py` are unlikely to remember to run it manually, so it'll drift out of sync with the implementation over time. It also sends an inaccurate signal to other contributors ("this repo welcomes tests"), which would create awkward conversations down the line.

The good news: the docstring you wrote on `_resolve_url` already enumerates the three URL forms clearly — code review alone is enough to verify correctness, which makes the test assertions redundant. So I'd like to ask you to drop the test file. **This isn't a comment on the test quality** (which is good), purely about keeping the repo's overall posture consistent.

**2. On the default endpoint switch — I'd prefer to flip it back**

Realistically, users who go out of their way to configure the MiniMax backend are overwhelmingly in mainland China. The original default `api.minimaxi.com` was specifically a zero-config path for them. Switching to `.io` means domestic users hit a connection failure on first run and have to dig through docs to add an env var before things work — the cost asymmetry doesn't favor the switch.

I'd suggest keeping `api.minimaxi.com` as the default and documenting `api.minimax.io` as an **override for overseas users**. That way:

- Domestic users (majority): zero config, just works
- Overseas users (minority): one extra line, `MINIMAX_BASE_URL=https://api.minimax.io`

**3. `.env.example` needs to be synced**

Regardless of which default we land on, `.env.example` L47 and L91 need to be updated to match — otherwise users following `.env.example` will end up with config that disagrees with the actual default.

---

**What I'd like to keep as-is:**

- ✅ `_resolve_url` fix for the `…/v1` edge case (the core value of this PR)
- ✅ MiniMax `.env` example in `docs/image.md`
- ✅ `MiniMax 2.7` → `MiniMax-M2.7` correction in README / README_CN

If those three points can be addressed, I'm happy to merge. Thanks again — genuinely nice work 🙏

PR #62, opened 2026-04-26
fix(svg_to_pptx): tolerate HTML named entities in AI-generated SVG

garvinwong:
## Problem

`xml.etree.ElementTree` only recognises the five XML reserved entities (`lt`, `gt`, `amp`, `quot`, `apos`). When an SVG contains an HTML named entity such as `&nbsp;`, `&mdash;`, or `&copy;`, parsing aborts with:

```
xml.etree.ElementTree.ParseError: undefined entity: line N, column M
```

This is a recurring issue with AI-generated decks — Claude / GPT / Gemini frequently emit `&nbsp;` inside `<text>`. A single offending slide currently kills the export of an entire deck. In real usage I just hit this on slide 4 of a 16-slide consulting-style PPT, and even though 15/16 slides converted fine, the whole `.pptx` was discarded.

## Fix

Pre-parse pass that replaces every **unknown** HTML named entity with its Unicode code point (`&nbsp;` → U+00A0, `&mdash;` → U+2014, etc.) by looking it up in `html.entities.html5`. The five XML reserved entities and numeric character references (`&#160;` / `&#xa0;`) pass through unchanged, and any genuinely unrecognised entity is left intact so the parser still surfaces a clear error if the input is malformed in some other way.

## Test

Minimal reproducer — fails on `main`, passes with this PR:

```python
from pathlib import Path
import tempfile
from svg_to_pptx import convert_svg_to_slide_shapes

svg = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">' \
      '<text x="10" y="50">a&nbsp;b&mdash;c</text></svg>'
with tempfile.NamedTemporaryFile(suffix='.svg', mode='w', delete=False) as f:
    f.write(svg); p = f.name
convert_svg_to_slide_shapes(Path(p))
```

I also re-ran the conversion against the original failing 16-slide deck after applying this patch and all 16 slides now convert.

## Scope

Single-file change, ~26 lines, no new dependency (uses stdlib `html.entities.html5`). Behaviour is strictly additive — every input that parsed before still parses identically.

### DIFF
--- a/skills/ppt-master/scripts/svg_to_pptx/drawingml_converter.py
+++ b/skills/ppt-master/scripts/svg_to_pptx/drawingml_converter.py
@@ -3,8 +3,31 @@
 from __future__ import annotations
 
 import re
+from html.entities import html5 as _HTML5_ENTITIES
 from pathlib import Path
 from xml.etree import ElementTree as ET
+
+# AI-generated SVGs frequently contain HTML named entities such as &nbsp;
+# &mdash; &copy;.  SVG is strict XML, so ElementTree only recognises the five
+# XML reserved entities (lt/gt/amp/quot/apos) and raises ParseError on the
+# rest.  This pre-parse pass replaces unknown HTML entities with their Unicode
+# code-point equivalents (e.g. &nbsp; -> U+00A0); the five XML reserved
+# entities and numeric character references (&#160; / &#xa0;) are kept
+# verbatim, and any unrecognised entity is left untouched so the parser can
+# still surface a clear error.
+_XML_RESERVED_ENTITIES = {"lt", "gt", "amp", "quot", "apos"}
+_ENTITY_RE = re.compile(r"&([#a-zA-Z][a-zA-Z0-9]*);")
+
+
+def _sanitize_svg_text(text: str) -> str:
+    def _replace(match: re.Match[str]) -> str:
+        name = match.group(1)
+        if name in _XML_RESERVED_ENTITIES or name.startswith("#"):
+            return match.group(0)
+        char = _HTML5_ENTITIES.get(name + ";") or _HTML5_ENTITIES.get(name)
+        return char if char else match.group(0)
+
+    return _ENTITY_RE.sub(_replace, text)
 
 from .drawingml_context import ConvertContext, ShapeResult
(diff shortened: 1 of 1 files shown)


hugohe3:
Thanks @garvinwong — clean diagnosis, clear reproducer, and the "one bad slide kills the whole deck" pain is real. I want to explain why I'm closing this rather than merging, and what landed instead.

**Why not at the export layer**

1. The sanitizer regex requires a trailing semicolon, so it covers `&nbsp;` / `&mdash;` but not bare `&` (`R&D`) or bare `<` / `>` (`error < 5%`). Same `ParseError`, same dead deck — 2 of 4 minimal repros still abort with this patch applied.
2. Absorbing malformed SVG at export silently relaxes the project's quality contract. PPT Master's value depends on the SVG being valid; once we sanitize at export we lose the signal that the model output was wrong.

**What landed on `main` instead**

The real gap was upstream — `svg_quality_checker.py` did no XML well-formedness check at all, and `shared-standards.md` only forbade *structural* features. A model could break XML without violating any documented rule.

- **`5b2686e`** — checker now runs `ET.fromstring` as the first gate, reporting offenders with file / line / column. Catches all four failure classes, not just two. Zero regressions on 228 existing examples.
- **`4b88398`** — `shared-standards.md` §1.0 now requires raw Unicode for typography (`—` `©` `→` NBSP …) and XML escapes for the five reserved characters, with the rule echoed in `CLAUDE.md` and the four reference docs that already enumerate SVG bans.

For your 16-slide scenario: the offending page is now caught before export with a precise location, and the model is less likely to emit it at all. Same end-state, upstream side, strict-XML invariant preserved.

Your PR forced the audit that surfaced both gaps — that's the real value here. Please keep them coming.

hugohe3:
Just to add a final note following up on this: after thorough investigation, the root cause is that the initial SVG generated by the AI is fundamentally corrupted as an XML document, which inevitably triggers these downstream parsing errors. The correct architectural fix here isn't to build tolerance for broken SVGs into our downstream tools, but rather to strictly enforce the generation of correct, valid SVGs right from the start. Thanks again for bringing this to light!

PR #83, opened 2026-05-02
feat(image-gen): add ModelScope backend for image generation

Ahusm:
feat(image-gen): add ModelScope backend for image generation

Introduces a new image generation backend using ModelScope (魔塔社区)

API with support for Z-Image-Turbo and other models.

Key features:

Async task-based generation workflow with polling

Multiple resolution presets (512px, 1K, 2K, 4K)

Aspect ratio support (1:1, 3:4, 4:3, 9:16, 16:9)

Configurable via MODELSCOPE_API_KEY, MODELSCOPE_MODEL, MODELSCOPE_BASE_URL

Configuration:

IMAGE_BACKEND=modelscope

MODELSCOPE_API_KEY=your-api-key

MODELSCOPE_MODEL=Tongyi-MAI/Z-Image-Turbo (default)

MODELSCOPE_BASE_URL=[https://api-inference.modelscope.cn/v1](vscode-file://vscode-app/d:/ProgramFiles/Microsoft%20VS%20Code/034f571df5/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (default)

Also updates .gitignore to exclude Python version files and build artifacts.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -80,14 +80,6 @@
 # VOLCENGINE_BASE_URL=https://operator.las.cn-beijing.volces.com/api/v1/images/generations
 
 # ─────────────────────────────────────────────────────────────
-# Example: ModelScope / 魔塔社区示例
-# ─────────────────────────────────────────────────────────────
-# IMAGE_BACKEND=modelscope
-# MODELSCOPE_API_KEY=your-modelscope-key
-# MODELSCOPE_MODEL=Tongyi-MAI/Z-Image-Turbo
-# MODELSCOPE_BASE_URL=https://api-inference.modelscope.cn/v1
-
-# ─────────────────────────────────────────────────────────────
 # Extended / Experimental backends
 # 扩展 / 实验后端
 # ─────────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


hugohe3:
I recommend not merging this PR yet. Adding ModelScope as an image backend is a reasonable direction and it fits as an experimental backend, but the current implementation does not meet the minimum behavior contract used by the existing image backends in this repository.

Blocking issues:

1. In `backend_modelscope.py`, the `SUCCEED` branch saves the image and then only `break`s; it never returns the saved path. The `FAILED` branch also only `break`s. As a result, `_generate_image()` can implicitly return `None`, and the CLI may exit successfully without producing a usable image.
2. Task polling uses an unbounded `while True`. If the remote task stays pending/running, the PPT generation flow can hang forever. The repo already has `backend_common.poll_json()` for ready/failed/timeout handling, so this should reuse it or implement equivalent total-timeout behavior.
3. Image download uses raw `requests.get(...).content` without timeout, HTTP status validation, or content-type/format handling. Existing backends consistently use `download_image()` or `save_image_bytes()`; this backend should do the same.
4. `_resolve_url()` uses `rstrip("/v1")`, which strips a character set rather than a suffix. It also forces the endpoint to end in `.cn`, which breaks `MODELSCOPE_BASE_URL` for proxies, test services, and compatible gateways.
5. `.gitignore` should not ignore `pyproject.toml`; that would block future project-level Python/uv configuration. `System.IO.Stream+NullStream` also looks like a local artifact and should not be part of this PR.

Suggested minimal fixes:

- On `SUCCEED`, validate that `output_images` is present, then `return download_image(image_url, path)`.
- On `FAILED`, raise a `RuntimeError` that includes the remote response payload.
- Use `poll_json(..., status_label="task_status", ready_values=["SUCCEED"], failed_values=["FAILED"])`, or add equivalent bounded polling with a total timeout.
- Use `download_image()` for the final image URL instead of reading `.content` directly.
- Normalize only the optional `/v1` suffix in the base URL; do not hard-code a domain allowlist.
- Remove the unrelated `.gitignore` entries.

Platform-wise, ModelScope is worth supporting, especially for Chinese prompts, domestic access, and the open-source/LoRA model ecosystem. However, its image API examples vary between synchronous and asynchronous response shapes, so this backend needs conservative handling for task state, timeouts, missing fields, and download failures. I would keep it experimental after these fixes and avoid promoting it to a core backend for now.

PR #93, opened 2026-05-08
feat(svg_to_pptx): add Japanese and Korean fonts to EA fonts list

yuki-katakami:
## Summary
- Add Japanese fonts (`Noto Sans JP`, `Yu Gothic`, `Meiryo`, `MS Gothic`, `Source Han Sans/Serif JP`, additional Hiragino variants) to the EA fonts set so they're correctly classified as East Asian
- Add Korean fonts (`Malgun Gothic`, `Gulim`, `Dotum`, `Batang`, `Noto Sans/Serif KR`)
- Add Windows fallback mappings for Japanese-specific aliases (`メイリオ` → `Meiryo`, `Source Han Sans/Serif JP` → `Noto Sans/Serif JP`)

## Motivation
Without these entries, Japanese/Korean text rendered with fonts like `Noto Sans JP` was not classified as East Asian, leading to incorrect typeface assignment in generated PPTX. The existing list focused on Simplified/Traditional Chinese fonts; this PR extends coverage to JP/KR to make the renderer usable in those locales without further changes.

## Test plan
- [x] Generated 23 Japanese-language slides locally with `Noto Sans JP` / `Meiryo` and verified correct East Asian font assignment in the resulting PPTX
- [ ] No new regressions for existing zh-CN / zh-TW handling (additive changes only)

### DIFF
--- a/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
+++ b/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
@@ -34,12 +34,23 @@ EA_FONTS = {
     'SimSun', 'SimHei', 'FangSong', 'KaiTi', 'STKaiti',
     'STHeiti', 'STSong', 'STFangsong', 'STXihei', 'STZhongsong',
     'Hiragino Sans', 'Hiragino Sans GB', 'Hiragino Mincho ProN',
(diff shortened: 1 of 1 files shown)


hugohe3:
Thanks @yuki-katakami — clean additive change, merged. The JP/KR coverage gap was real; this makes the renderer usable out-of-the-box in those locales.

PR #107, opened 2026-05-16
feat(image): reattach local_workbench and align size mapping with loc…

zl911999111:
## What
Reattach Hanako's local image workbench backend to the latest `image_gen.py` pipeline and align its size mapping with the current local image service.

## Changes
- add `local_workbench` backend registration back into `skills/ppt-master/scripts/image_gen.py`
- add `skills/ppt-master/scripts/image_backends/backend_local_workbench.py`
- ignore local `.venv/` in `.gitignore`
- align local workbench aspect-ratio-to-size mapping with the current `image2` local service config

## Why
The local image backend had diverged from upstream during recent `ppt-master` updates.

Two issues needed to be resolved:
1. the local backend was no longer attached to the latest unified image generation pipeline
2. the old hardcoded size mapping no longer matched the current local image service, which caused upstream image requests to fail

## Validation
Verified locally:
- `image_gen.py --help` still exposes the latest manifest-related flags
- `local_workbench` is available in backend listing
- single-image generation works through `local_workbench`
- manifest mode works and writes status back correctly
- markdown sidecar generation still works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,6 +113,7 @@ __pycache__/
 *$py.class
 *.so
 .Python
+.venv/
 env/
 venv/
 ENV/
--- a/skills/ppt-master/scripts/image_backends/backend_local_workbench.py
+++ b/skills/ppt-master/scripts/image_backends/backend_local_workbench.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+"""
+PPT Master backend that delegates generation to Hanako's local GPT Image 2 workbench.
+
+Persistent source:
+  D:/HanakoField/persistent-overrides/ppt-master-local-image/backend_local_workbench.py
+
+Runtime copy installed into:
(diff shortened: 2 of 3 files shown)


hugohe3:
@codex review

hugohe3:
I'm closing this PR.

This PR is not acceptable. This is not your personal project, and you should not submit your own private image reverse-proxy/workbench into the main repository as if it were a supported platform.

The implementation hardcodes a machine-specific path:

D:/HanakoField/image2本地使用/openai_compatible_image.py

It depends on your private local environment, private .env, private script, and API behavior that the project cannot review or reproduce. Registering this as a core backend is especially wrong, because it makes a personal setup look like an officially supported provider.

A backend contribution must be generic, configurable, reproducible, documented, and safe for other users. This PR is none of those.

Closing.

PR #124, opened 2026-05-21
fix(svg-editor): backfill SVG dimensions from viewBox to fix blank live preview

### LINKED ISSUE #115: 实时预览不显示
### Problem area

Other

### What happened?

您的项目做的太好了，用起来很顺手。但我在使用过程中，发现实时预览无法显示，尝试不同项目和浏览器，还是一样，想请教下您这是什么原因呢？
Image

### Steps to reproduce

no 

### Expected behavior

_No response_

### AI editor / tool

claude code

### AI model

_No response_

### OS

_No response_

### Python version

_No response_

### Screenshots or files

_No response_

jysir99:
### Root Cause
`sanitizeSvg()` in `app.js` uses `DOMParser` → `XMLSerializer` round-trip to clean incoming SVG. This round-trip strips `width`/`height` attributes when they are implicit (derived from `viewBox`). The browser then renders the SVG at 0x0, causing a blank preview.

### Fix
Backfill `width` and `height` from `viewBox` after parsing, only when the attributes are missing.

### Testing
- Verified with 2-page SVG deck (latex_formula_demo project) — live preview renders correctly with and without explicit width/height on source SVGs.
- No changes to shared SVG generation standards; fix is entirely frontend-side.

Closes #115

### DIFF
--- a/skills/ppt-master/scripts/svg_editor/static/app.js
+++ b/skills/ppt-master/scripts/svg_editor/static/app.js
@@ -354,6 +354,18 @@
                 // Empty-canvas guard: surface a clear error if the SVG parsed
                 // to nothing renderable (issue #115's silent-blank scenario).
                 var rootSvg = svgContent.querySelector("svg");
+                // DOMParser → XMLSerializer round-trip strips implicit dimensions,
+                // collapsing the SVG to 0×0 in the browser. Backfill from viewBox.
+                if (rootSvg && !rootSvg.hasAttribute("width")) {
+                    var vb = rootSvg.getAttribute("viewBox");
+                    if (vb) {
+                        var parts = vb.trim().split(/[\s,]+/);
+                        if (parts.length === 4) {
+                            rootSvg.setAttribute("width", parts[2]);
+                            rootSvg.setAttribute("height", parts[3]);
+                        }
+                    }
+                }
                 var hasContent = false;
                 if (rootSvg) {
                     var children = rootSvg.querySelectorAll("*");


hugohe3:
Thanks for the PR 🙏 Not merging this one — laying out the reasons together.

**1. Root cause description is inaccurate**

The PR attributes the bug to `sanitizeSvg()`'s `DOMParser → XMLSerializer` round-trip, but that step does not strip existing attributes, nor does it synthesize `width`/`height` from `viewBox`. The real cause is that the input SVG itself lacks `width`/`height` — @USTCTaoXu had already pinpointed it in the #115 thread: "SVG root only has viewBox, no width/height attributes".

**2. The project has a mandatory spec for SVG output**

`references/shared-standards.md:172`:

> viewBox must match the canvas dimensions (`width`/`height` must match `viewBox`)

The repo's own `finalize_svg` / Executor output always emits `viewBox + width + height` together. The normal pipeline never produces dimensionless input. The SVG in #115 was missing `width`/`height` — that's the upstream output violating the spec.

**3. The real disagreement is about which direction to fix in (this is the core point)**

The upstream spec is mandatory. Downstream is not here to keep adapting to outputs that ignore the spec.

An analogy: if the previous stage of a workflow keeps producing garbage, should the next stage keep cleaning up after it? Of course not. The standard flow is — whoever produces the mess cleans it up. The fix doesn't belong on the downstream side patching holes because the responsible party didn't do their job.

This project has followed that logic from the start: walk along the correct path, instead of treating the symptom wherever it shows up (头疼医头,脚疼医脚 — "treat the head when the head hurts, treat the foot when the foot hurts"). When upstream violates the spec, the correct action is to fix upstream, or — like the observability layer I just added on #115 — surface the error explicitly so the producer fixes it. Not to insert a silent absorption layer downstream that makes non-compliant input "just work". Once that kind of adaptation starts there is no boundary: every spec-violating input demands a new fallback, downstream eventually fills up with unrelated compensation code, and the upstream responsibility gets diluted.

The Empty-canvas guard right next to this code, and the observability added in #115, are both on the same surface-error route. This PR's silent backfill cuts straight across that route — the direction is reversed.

**4. Issue #115 is already resolved**

After pulling the new main and seeing the explicit error signal, @USTCTaoXu fixed it on the server side by filling in `width`/`height` and closed the issue themselves — that's the spec-compliant fix. A frontend fallback isn't needed on top of it.

Thanks again for digging into this 🙏

jysir99:
Thanks for the detailed review — I agree with all four points.

I have updated this PR to remove the silent backfill entirely. The new approach:

1. **No backfill** — SVG is not modified in any way.
2. **Visible error banner** — when the live preview detects a root svg without width/height, it displays a centered red card telling the user: "Please ask the AI to strictly follow shared-standards.md §4 and include width & height in the SVG root element."
3. **Per-slide lifecycle** — the banner is removed on navigation; it only re-appears if the next slide also lacks dimensions.

This aligns with the "surface the error" direction you described. The upstream spec stays authoritative; the frontend just makes spec violations visible instead of silently blank.

Would love your thoughts on this approach.

PR #127, opened 2026-05-22
fix(svg-editor): show spec-violation banner when SVG lacks width/height

### LINKED ISSUE #115: 实时预览不显示

jysir99:
Closes #115. Surface spec violations with a red banner instead of silent blank preview. No backfill - upstream spec stays authoritative.

### DIFF
--- a/skills/ppt-master/scripts/svg_editor/static/app.js
+++ b/skills/ppt-master/scripts/svg_editor/static/app.js
@@ -38,6 +38,7 @@
             err_save: "Save failed: ",
             err_empty_svg: "Slide loaded but the canvas is empty. The SVG may be malformed or missing a root <svg> element.",
             warn_icon_inline: "{count} icon(s) failed to render: {names}",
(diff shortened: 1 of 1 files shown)


hugohe3:
Thanks for the contribution. I think the direction here is right: live preview should not silently backfill root SVG dimensions on the frontend, because that would hide upstream SVG output that does not follow shared-standards.md §4. Surfacing the missing width/height as a spec violation is a better behavior than continuing to fail silently.

I made one small follow-up fix before merging: the original implementation only checked for a missing width attribute, so it missed the case where height was absent. The merged version now triggers the warning when either width or height is missing.

PR #129, opened 2026-05-23
feat(latex): add LaTeX formula rendering via CodeCogs API

jysir99:
## Summary

Add LaTeX formula rendering support to the PPT generation pipeline. Formulas written in `spec_lock.md` using `$...$` (inline) or `$$...$$` (block) syntax are automatically rendered to transparent PNG images via the CodeCogs online LaTeX API.

### Before / After

Snipaste_2026-05-23_10-19-13
Snipaste_2026-05-23_10-19-44

---

### Changes

**New: `scripts/latex_render.py`**
- Pure CodeCogs API backend — no local LaTeX installation required
- Extracts `$...$` and `$$...$$` formulas from `spec_lock.md`
- Renders each unique formula to `images/latex/F*.png`
- Generates `images/latex/manifest.json` with:
  - `uniform_scale`: single scale factor based on tallest formula → target display height (50px), ensuring letter sizes stay consistent across all formulas
  - Per-formula dimensions and type (inline/block)
- Supports `--dry-run` (preview without rendering) and `--dpi` (default 300)

**Modified: `SKILL.md`**
- Added Step 4.5 between Step 4 and Step 5 for the LaTeX rendering step

**Modified: `references/strategist.md`**
- LaTeX formula identification rule: formulas are pre-rendered PNG images, not text rendered with any font stack
- Typography System chapter clarifies formula images align with surrounding typography via `manifest.json` `uniform_scale`

**Modified: `references/executor-base.md`**
- Formula images placed as `<image>` elements using `uniform_scale` from manifest
- Clarified formulas are not a font-stack role

**Modified: `templates/spec_lock_reference.md`**
- `code_family` applies to code snippets only; LaTeX formulas are pre-rendered PNG images

**Modified: `CLAUDE.md`**
- Added `latex_render.py` commands to Command Quick Reference

### Key Design Decision: `uniform_scale`

All formula images share a single scale factor rather than being normalized to the same height individually. Normalizing heights would break letter size consistency — a short formula like $E=mc^2$ would have its letters enlarged to match the tallest formula's height, making it visually inconsistent with surrounding text. The uniform scale factor preserves the original relative sizes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,10 @@ python3 skills/ppt-master/scripts/total_md_split.py <project_path>
 python3 skills/ppt-master/scripts/finalize_svg.py <project_path>
 python3 skills/ppt-master/scripts/svg_to_pptx.py <project_path>
 # Add --merge-paragraphs when the user wants paragraph-level editable text frames instead of one-per-line (default off, see SKILL.md Step 7.3).
+
+# Formula rendering (only when spec_lock.md contains LaTeX)
+python3 skills/ppt-master/scripts/latex_render.py <project_path>
+python3 skills/ppt-master/scripts/latex_render.py <project_path> --dry-run
 ```
 
 ## Core Directories
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,6 +82,7 @@ python3 skills/ppt-master/scripts/svg_to_pptx.py <project_path>
 # Formula rendering (only when spec_lock.md contains LaTeX)
 python3 skills/ppt-master/scripts/latex_render.py <project_path>
 python3 skills/ppt-master/scripts/latex_render.py <project_path> --dry-run
+python3 skills/ppt-master/scripts/latex_render.py <project_path> --dpi 400
 ```
 
 ## Core Directories
--- a/skills/ppt-master/SKILL.md
+++ b/skills/ppt-master/SKILL.md
@@ -291,6 +291,24 @@ python3 ${SKILL_DIR}/scripts/analyze_images.py <project_path>/images
 
 ---
 
+### Step 4.5: Formula Rendering (Conditional)
+
+🚧 **GATE**: Step 4 complete; `spec_lock.md` generated.
+
+> **Trigger**: `spec_lock.md` contains `$...$` or `$$...$$` LaTeX formulas. If no formulas are detected, skip to Step 5.
+
+```bash
+python3 ${SKILL_DIR}/scripts/latex_render.py <project_path>
+```
+
+This scans `spec_lock.md`, renders each LaTeX formula to a transparent PNG image in `images/latex/`, and replaces the formula text with `![](images/latex/F01.png)` references. Executor treats these as normal images — no Executor changes needed.
+
+**Dry-run available**: `python3 ${SKILL_DIR}/scripts/latex_render.py <project_path> --dry-run` shows what would be rendered without modifying files.
+
+**✅ Checkpoint — Formulas rendered (or no formulas found). Auto-proceed to Step 5.**
+
+---
+
 ### Step 5: Image Acquisition Phase (Conditional)
 
 🚧 **GATE**: Step 4 complete; Design Specification & Content Outline generated and user confirmed.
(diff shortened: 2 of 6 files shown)


hugohe3:
Thanks for the contribution — really appreciate the thought that went into it. I want to evaluate this carefully before deciding, so I'm going to hold off on merging for now.

My read on what's being proposed: the core capability is **LaTeX formula handling in decks**, and the implementation path is "formula → transparent PNG → embedded as `<image>` in SVG". Let me know if that misses anything.

A few things I want to think through before deciding:

1. **Pipeline placement** — should formula rendering live in the main pipeline as a mandatory conditional step (current PR's Step 4.5), or sit alongside the main flow as an optional capability? Most decks don't contain formulas, and where this lives affects the cost model for the common case.
2. **Upstream normalization** — should the Strategist be responsible for normalizing math expressions in source documents to LaTeX before downstream consumption? That decides where the contract lives and how robust the input side is.
3. **External API stability** — the implementation depends on CodeCogs as a single rendering source. I want to think through what happens when it's unavailable, rate-limited, or changes policy.

This may well be a direction the repo grows into. But I'd rather get the path right before merging — going to take some time on this and circle back. Thanks for your patience.

hugohe3:
Hi @jysir99 — thanks for putting this together. The screenshots show exactly the kind of result formula-heavy decks need, and this PR is what finally pushed LaTeX rendering out of the roadmap's "Under consideration" section. The shipped implementation owes its existence to the question this PR asked.

I've ended up **not merging this PR as-is** and instead reshaping the direction around a few existing project axioms. Walking through where the two implementations diverged so the reasoning is visible, not just the outcome:

**1. Formula selection: explicit Strategist manifest, not a regex scan of `spec_lock.md`**

`spec_lock.md` is a single-source-of-truth artifact written by the Strategist during the lock step — nothing downstream is supposed to edit it. The regex-find-and-replace pattern (`$...$` → `` written back into `spec_lock.md`) makes the renderer a downstream writer to upstream truth, which is a pattern the rest of the pipeline actively avoids (design_spec and spec_lock are written by the same AI in one pass; no sync scripts).

The shipped version reads `images/formula_manifest.json`, written by the Strategist after typography confirmation. *Which* expressions become PNGs is an AI judgment under a user-chosen policy (`mixed` / `render-all` / `text-only`), not a regex outcome.

**2. Multi-provider fallback chain, not CodeCogs-only**

CodeCogs is the right primary, but a single public endpoint is a single point of failure. The renderer walks `codecogs → quicklatex → mathpad → wikimedia` and records the winning provider in the manifest. None of the four require API keys.

**3. Native ratio per formula, not `uniform_scale`**

`uniform_scale` is a clean idea, but normalizing every formula's height to the tallest one forces inline expressions like `O(n log n)` to letter-match a multi-line integral, which actually inverts the visual hierarchy in formula-heavy decks. The manifest now carries `pixel_width / pixel_height / ratio` per item, and the Executor uses `preserveAspectRatio="xMidYMid meet"` per slot — different visual weight maps to different rendered sizes by design.

**4. Step 4 typography item, not Step 4.5**

The Eight Confirmations are a load-bearing structural commitment — users see the same eight items plus a mode line every run. Splitting the pipeline at 4.5 breaks that shape. The formula policy folded into typography confirmation (item 7) instead, with three explicit options the user picks or overrides.

**5. First-class image rows, not a side-channel asset**

Formula PNGs flow through the same Image Resource List as every other image: `Acquire Via: formula` / `Status: Rendered` / `Type: Latex Formula` in §VIII, with matching entries in `image-base.md`, `svg-image-embedding.md`, `executor-base.md`, and the spec_lock reference. The Executor handles them via the same `no-crop` / `preserveAspectRatio="xMidYMid meet"` path used for charts and screenshots.

The shipped version landed in 920ac1d0 on `main`.

### Where contributions on the shipped version would be welcome

- Additional no-key public renderers for the fallback chain
- Provider-specific quality fixes (e.g., wikimedia doesn't preserve formula color — currently flagged via a `color_warning` field but not corrected)
- A small corpus of test formulas for regression testing across providers

Closing this PR — thanks again for the push, the shipped feature wouldn't have landed this week without it.

hugohe3:
Closing — superseded by 920ac1d0 on main. See comment above for the design walk-through. Thanks again @jysir99.

hugohe3:
And to make the invitation explicit — the three items above are starting points, not a task list. Any follow-up PR against the shipped version is welcome, including ones that revisit the design choices in the walk-through if you have a use case where the trade-off cuts the other way.

PR #130, opened 2026-05-23
fix(svg): require width+height+viewBox on SVG root element

jysir99:
## Summary

SVG root elements must include `width`, `height`, and `viewBox` — all three attributes are required. Omitting `width`/`height` causes browser preview and some SVG renderers to display slides at incorrect size.

### Changes

**Modified: `references/shared-standards.md`**
- Updated §4 SVG rule: `viewBox must match canvas dimensions` → explicit requirement that all three attributes (`width`, `height`, `viewBox`) must be present with identical values

**Modified: `scripts/svg_quality_checker.py`**
- Added missing `SVG_NS = "http://www.w3.org/2000/svg"` constant used for namespace-aware XML parsing

### DIFF
--- a/skills/ppt-master/references/shared-standards.md
+++ b/skills/ppt-master/references/shared-standards.md
@@ -169,7 +169,7 @@ One offending character invalidates the file and aborts export. Numeric refs (`&
 
 ## 4. Basic SVG Rules
 
-- **viewBox** must match the canvas dimensions (`width`/`height` must match `viewBox`)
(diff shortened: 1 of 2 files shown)


hugohe3:
Thanks for the patch — closing this without merging, with context:

**1. `SVG_NS` constant** — already landed on main in `36cd1571` (2026-05-22, one day before your branch). Your branch forked from `60712cd3`, so the existing fix didn't show in your diff.

**2. Rule expansion in `shared-standards.md`** — I'd like to keep this one terse. The existing line ("`width`/`height` must match `viewBox`") already implies both must be present, since "match" requires existence. PPT Master's rule set is already long, and every line is a token the AI has to budget against the rest of the prompt stack. I only expand a rule when there's a recurring failure mode reported against generated decks — a one-off renderer-preview quirk doesn't justify the prose budget yet.

If you hit decks where the AI actually omits `width`/`height` from the `<svg>` root and the deck breaks in PowerPoint, please file an issue with the repro and I'll revisit.

Thanks again for taking the time.

PR #146, opened 2026-06-03
fix(pptx_to_svg): resolve font-size from pPr/defRPr in PPTX→SVG text export

ZhaoZuohong:
## Problem

`_build_run()` in `txbody_to_svg.py` only reads `sz` (font size) from `<a:rPr>` (run-level) and `<a:endParaRPr>` (end-of-paragraph marker). When PPTX text runs lack explicit `rPr@sz` because the font size is defined at the paragraph level (`<a:pPr><a:defRPr sz="2800">`), the hardcoded `DEFAULT_FONT_SIZE_PX` (24px = 18pt) was used as fallback.

**Result**: imported SVG shows wrong font sizes for all inherited text.

## How to reproduce

1. Create a PPTX with a text placeholder where font size is set via the placeholder's default text style (e.g., title at 28pt, body at 24pt), not on individual text runs.
2. Run `pptx_template_import.py` on it.
3. The flat SVG shows `font-size="24"` (18pt) for ALL text, regardless of the actual placeholder font size.

## Fix

- Adds `defRPr` to the `_attr_chain` in `_build_run()`, with correct DrawingML priority: **rPr → defRPr → endParaRPr**
- Applies the same fallback chain to `spc` (letter-spacing), font typeface resolution, and color source resolution
- Threads a `default_font_size_px` parameter through the full stack (`convert_txbody` → `_parse_paragraphs` → `_parse_paragraph` → `_build_run`) so callers can pass placeholder-aware defaults in the future
- Callers in `slide_to_svg.py` pass `DEFAULT_FONT_SIZE_PX` explicitly, preserving existing behavior

## Changes

| File | + | - |
|------|---|---|
| `pptx_to_svg/txbody_to_svg.py` | +30 | -17 |
| `pptx_to_svg/slide_to_svg.py` | +3 | -0 |

## Verification

- `pptx_template_import.py` runs without errors on test PPTX
- Parameter chain verified via `inspect.signature(convert_txbody)`

### DIFF
--- a/skills/ppt-master/scripts/pptx_to_svg/slide_to_svg.py
+++ b/skills/ppt-master/scripts/pptx_to_svg/slide_to_svg.py
@@ -44,6 +44,7 @@ from .txbody_to_svg import (
     convert_txbody,
     convert_vertical_txbody,
     is_vertical_txbody,
+    DEFAULT_FONT_SIZE_PX,
 )
 
 
@@ -321,6 +322,7 @@ def _convert_shape(node: ShapeNode, ctx: AssemblyContext, *, top_level: bool) ->
             tx_body, node.xfrm, ctx.palette,
             theme_fonts=ctx.theme_fonts,
             default_fill=text_default_fill,
+            default_font_size_px=DEFAULT_FONT_SIZE_PX,
             id_prefix=f"{ctx.group_id_prefix}txt",
             id_seq=ctx.grad_seq,
         )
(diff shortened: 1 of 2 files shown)


hugohe3:
Merged. This fixes the paragraph-local pPr/defRPr fallback path and keeps the run-level override behavior intact. I found a related remaining gap in txBody lstStyle/lvlNpPr/defRPr and placeholder style inheritance, so I will follow up with a separate fix on top of this merge rather than widening this PR after the fact.

hugohe3:
Thanks for the focused fix here. This solves an important part of the text-style fallback issue: paragraph-local pPr/defRPr is now respected, while run-level rPr still takes precedence. It does not cover the full placeholder/default-style inheritance chain yet, especially txBody lstStyle/lvlNpPr/defRPr and layout/master placeholder text styles. I am continuing with a follow-up fix for those remaining cases on top of this merge.

PR #150, opened 2026-06-07
fix(svg_to_pptx): treat digits as tabular width in estimate_text_width

PHY041:
## What

`estimate_text_width()` classes the digit `1` with the narrow glyphs (`iIlj!|`, 0.3 em) and lets `%` fall into the default 0.55 em bucket. In the UI fonts the converter targets (Trebuchet MS, Calibri, Segoe UI, Microsoft YaHei), digits are **tabular** — all ≈0.55 em wide including `1` — and `%` is a wide glyph (≈0.75 em).

This PR moves digits to a 0.55 em class and `%` to the wide class. 7 insertions, 2 deletions, no behavior change for non-digit text.

## Why it matters

Any short label containing `1` gets an under-sized text box. Example at 58 px bold:

| Text | Old estimate | New estimate | Actually rendered (Trebuchet MS) |
|---|---|---|---|
| `12%` | 85.3 px | 112.7 px | ~110 px |

PowerPoint hides the bug because the exported boxes are `wrap="none"` + `spAutoFit` — the text just overflows the too-small extent. But **LibreOffice Impress ignores `wrap="none"`** and wraps the overflow, so decks break visibly there:

- `12%` renders as `12` / `%` on two lines
- KPI labels `P1`, node numbers `01`, prices `$99–299` all wrap the same way
- `94%`, `P0`, `+30%` render fine — which makes the bug look random until you notice every broken string contains a `1`

## Reproduction

1. Create a page with `<text font-family="Trebuchet MS" font-size="58" font-weight="700">12%</text>`
2. Export via `svg_to_pptx.py`
3. Open the native pptx in LibreOffice Impress → `12%` wraps onto two lines; PowerPoint renders fine

## Local verification

Re-exported a real 13-page deck containing `12%`, `+41%`, `+31%`, `01`, `P1`, `$99–299` labels:
- before: all of the above wrapped in LibreOffice (PowerPoint OK)
- after: all render on one line in both LibreOffice and PowerPoint; remaining pages byte-identical behavior

Also note for anyone tempted by the obvious workaround: padding the string with a trailing NBSP does **not** work — the converter trims it before measuring.

### DIFF
--- a/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
+++ b/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
@@ -449,10 +449,15 @@ def estimate_text_width(text: str, font_size: float, font_weight: str = '400') -
             width += font_size
         elif ch == ' ':
             width += font_size * 0.3
-        elif ch in 'mMwWOQ':
+        elif ch in 'mMwWOQ%':
             width += font_size * 0.75
-        elif ch in 'iIlj1!|':
+        elif ch in 'iIlj!|':
             width += font_size * 0.3
+        elif ch.isdigit():
+            # digits are tabular (uniform ~0.55em) in most UI fonts, including
+            # '1' — classing it with 'il|' under-sizes the box and makes
+            # renderers that ignore wrap="none" (LibreOffice) wrap the line
+            width += font_size * 0.55
         else:
             width += font_size * 0.55
 


hugohe3:
Thanks for the sharp catch and the thorough write-up — the LibreOffice repro and the "every broken string contains a 1" observation made the root cause obvious. Merged into `main`.

One follow-up I applied on top while merging: the same width heuristic is duplicated in the reverse direction at `pptx_to_svg/txbody_to_svg.py::_char_width`, which carries a comment stating it must mirror `estimate_text_width`. Your fix only touched the `svg_to_pptx` side, so I synced the mirror in [fb6a2a2c](https://github.com/hugohe3/ppt-master/commit/fb6a2a2c) — '1' moved to the tabular-digit class (0.55em) and '%' to the wide class (0.75em) there too — so the PPTX→SVG wrapping decisions stay aligned with the box-sizing estimate.

Both commits are now on `main`. Thanks again!

PR #154, opened 2026-06-10
feat(image-gen): add dedicated Agnes AI backend

Hann1balLecter:
Add backend_agnes.py for agnes-image-2.1-flash (OpenAI-compatible API without response_format=b64_json). Register in image_gen.py under IMAGE_BACKEND=agnes with AGNES_API_KEY / AGNES_BASE_URL / AGNES_MODEL env vars.

### DIFF
--- a/skills/ppt-master/scripts/image_backends/backend_agnes.py
+++ b/skills/ppt-master/scripts/image_backends/backend_agnes.py
@@ -1,0 +1,207 @@
+#!/usr/bin/env python3
+"""
+Agnes AI image generation backend.
+
+OpenAI-compatible API (agnes-image-2.1-flash etc.), but without
+response_format=b64_json — Agnes returns URLs only.
+
+Configuration keys:
+  AGNES_API_KEY    (required)
+  AGNES_BASE_URL   (optional, default: https://apihub.agnes-ai.com/v1)
+  AGNES_MODEL      (optional, default: agnes-image-2.1-flash)
+"""
+
+import sys
+
+if __name__ == "__main__" and any(arg in {"-h", "--help", "help"} for arg in sys.argv[1:]):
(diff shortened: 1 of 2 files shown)


hugohe3:
Thanks for the PR, and for surfacing Agnes AI compatibility!

We took a slightly different route rather than merging a dedicated `backend_agnes.py`. Agnes is an OpenAI-compatible endpoint whose only quirk is rejecting some OpenAI-specific request fields (`response_format`, `quality`, fixed size presets). Adding a whole per-provider backend for that would mean a new backend file for every OpenAI-compatible vendor with the same quirk, which we'd rather not proliferate.

Instead, in [`2b337bec`](https://github.com/hugohe3/ppt-master/commit/2b337bec18c2cb1381fee2dfca2e6f6fed463ce6) we generalized the existing `openai` backend with three optional compatibility knobs:

```env
IMAGE_BACKEND=openai
OPENAI_API_KEY=your-agnes-key
OPENAI_MODEL=agnes-image-2.1-flash
OPENAI_BASE_URL=https://apihub.agnes-ai.com/v1
OPENAI_SIZE_PRESET=gpt-image-2
OPENAI_RESPONSE_FORMAT=omit
OPENAI_QUALITY=omit
```

- `OPENAI_RESPONSE_FORMAT=omit` drops the field; the read side already handles URL responses, so it works end to end.
- `OPENAI_QUALITY=omit` drops the quality field.
- `OPENAI_SIZE_PRESET` decouples the size-mapping table from the model name.

This covers Agnes AND any other OpenAI-compatible provider with the same constraints, with no new backend to register. The Agnes example is documented in `.env.example` and `scripts/docs/image.md`.

Closing in favor of that approach — but genuinely appreciate the report. 🙏

PR #156, opened 2026-06-10
feat(native-diagram): component format + resolver core

shenqili:
Introduce the minimal native-diagram path: a pre-extracted DrawingML
component is spliced into a deck at SVG->PPTX time via a
`data-native-diagram` placeholder rect, arriving pixel-faithful and
natively editable (unlike the SVG-redraw path).

What's in this PR
- native_diagrams/component.py: the component format + theme-flattening
  primitives (schemeClr->srgbClr, theme-font tokens->typeface) so a lifted
  component renders identically in any deck.
- svg_to_pptx/native_diagram_resolver.py: the runtime splice path. Reads the
  stored <a:diagram> wrapper, scales it into the placeholder rect (EMU),
  remaps media rIds, renumbers <p:cNvPr> ids, and rewrites the wrapper to
  <p:grpSp> keeping namespace prefixes byte-exact.
- svg_to_pptx/drawingml_converter.py: intercept data-native-diagram before
  the normal tag dispatch.
- references/native-diagrams.md: selection + core placement spec.

The recolor / text-fill / font transforms layer on top in the stacked
text-font PR; this PR is splice-only.

Review fixes folded in
- resolver: fail loudly on a component key that escapes the library dir;
  warn (not silently drop) on a zero-size placeholder rect.
- converter: re-raise SvgNativeConversionError as-is instead of
  double-wrapping it.
- native-diagrams.md: reconcile the style-gate wording with the asset
  README/index (no hard gate, a soft cohesion call); drop a decorative
  emoji; cross-link the resolver script (docs/rules code-style 14).

Scope: base main, standalone; foundation for the text-font and engine PRs.
Follows docs/rules/. No demo decks / images / assets.
Verified: python -m py_compile on all changed modules.

---
1. #156 core

### DIFF
--- a/skills/ppt-master/references/native-diagrams.md
+++ b/skills/ppt-master/references/native-diagrams.md
@@ -1,0 +1,117 @@
+# Native Diagrams — Selection & Placement Reference
+
+> Lazy-loaded reference. Load **only** when `spec_lock.md page_diagrams` has at
+> least one entry (Strategist matched a page to a native diagram). If that section
+> is empty/absent, this file is never read. Peer to [`image-generator.md`](./image-generator.md)
+> and the chart catalog — same "scan an index, match by content shape" pattern.
+
+A **native diagram** is a pre-designed, fully-editable DrawingML figure lifted
+verbatim from a real deck and stored in `templates/native_diagrams/`. Unlike a
+`templates/charts/` SVG (which Executor *redraws* from scratch using the template
+as reference), a native diagram is **spliced in as-is** via a placeholder and only
+recolored — so an elaborate, polished figure arrives pixel-faithful and still
+native-editable, at near-zero authoring cost.
+
+---
+
+## 1. When a native diagram is the right vehicle (vs the alternatives)
+
+For any page whose content is a *structural relationship* (hierarchy / flow /
+convergence / comparison / composition / cycle), there are four vehicles. Pick by
+content-shape **and** the deck's visual ambition (set once in §d Style Objective):
+
+| Vehicle | Use when |
+|---|---|
+| Native diagram | The relationship maps to a polished, dimensional figure. Spliced in as-is + recolored to the deck palette — any deck can use it once recolored; keep the treatment consistent across the deck's structural pages. |
+| `charts/` SVG template | The deck is flat / minimalist, OR the figure needs structural surgery (different slot count, custom axes) the native diagram can't flex to. |
+| Hand-drawn SVG | No catalog entry fits; the page wants a bespoke layout. |
+| AI image (`image-generator`) | The page wants atmosphere / a scene / a hero, not an editable structural figure. |
+
+> **Visual-cohesion guardrail**: Native diagrams carry a deliberate **3D /
+> dimensional idiom** — recolor adapts their *color* to the deck palette, but the
+> glossy/volumetric *treatment* stays. There is **no hard style gate** (recolored,
+> they drop into any deck); it is a soft cohesion call — once you use them, use them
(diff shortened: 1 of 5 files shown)


hugohe3:
@shenqili First, genuine appreciation — I went through the implementation, and it's expert-level OOXML work. The theme-flattening design especially: resolving `schemeClr`→`srgbClr` while keeping the `lumMod`/`shade`/`tint` children intact, so a single base-hex remap re-derives every gradient and 3D tint — that's a clever core. The byte-exact namespace-prefix splice (string-based rather than ElementTree, so `a14`/`a16` extension prefixes aren't rewritten) and the foreign-rel stripping that keeps dangling rIds out of the target deck both show you know exactly where this gets dangerous. This is real craft.

That's also why I'd rather do it justice than rush it. I'd like to treat the five native-diagram PRs (#156, #158, #159, #160, #161) as one direction and settle the direction itself in an issue before reviewing them as code. Three things I'd want to work through first:

**1. Whether this belongs in the core pipeline.**
I'm not yet convinced the native-diagram path needs to live in the core generation flow. I deliberately did *not* take the lifted-diagram route when building the tool, because hand-authored SVG is already visually good enough for the structural figures we generate. I'm very open to being shown real cases where the SVG path genuinely falls short — that's exactly the kind of thing worth establishing in discussion, with concrete examples, rather than settling inside a five-PR implementation.

**2. Copyright / licensing — a precondition, not a side concern.**
This tooling is most useful precisely when there's a library of reusable DrawingML components behind it — and that library is, by design, lifted verbatim from existing third-party `.pptx` files (you've flagged this yourself by holding it "pending a licensing pass" and gitignoring the client-derived components). So the first thing we'd need to settle is what license-clean source that library can come from. For a repository this widely used, redistributing diagrams extracted from other people's decks — or shipping a default workflow that encourages it — is a legal exposure I have to avoid. Right now the machinery can merge, but no license-clean component library ships with it, so it has nothing it can actually use.

**3. Project positioning and recurring value.**
PPT Master builds the main *framework* of a deck, not the entire finished PPT end to end — users are expected to keep refining the output themselves. Within that boundary, I'm not yet convinced this automation provides enough recurring value to justify becoming core-pipeline machinery: a placeholder that pulls in a component, then needs its slots mapped by hand from `meta.json` in document order, sits fairly close to a user dropping in a suitable diagram and editing it directly (tools like OfficePlus or iSlide already serve that swap-in need well). That's a question about scope, not a knock on the work.

So — could we move this to an issue? I'd order the discussion as: (1) where the existing SVG path actually falls short, with real examples; (2) whether there's a license-clean, sustainable source for the component library; (3) whether this belongs in the core pipeline or as an external / experimental tool; and (4) if the direction holds, a single minimal end-to-end PR rather than the whole ecosystem at once. I'd rather not merge the implementation ahead of that conversation.

The engineering here is genuinely strong — thank you for the depth of work. Let's find the right shape for it together 🙏

hugohe3:
Hi @shenqili — thanks again for all the work that went into this, and for the thorough write-ups, the comparison table, and the runnable demo.

As we've talked through across the earlier threads (#167 and the related discussions), the native-diagram direction isn't something I want to push forward at this stage. As we discussed, this is about it sitting outside the project's current focus rather than being a problem with your implementation.

To keep the PR list manageable, I'm going to close this set as not planned for now, rather than leave it open indefinitely. This isn't a hard "no" to the idea itself — when the timing is right and we decide to revisit this direction, we can reopen and restart from here.

Really appreciate the effort and your understanding.

PR #157, opened 2026-06-10
docs(visual-design): premium paradigm + exemplars

shenqili:
The premium hand-authored-SVG design system and its feedback-evolution
(exemplars) mechanism. Independent of the native-diagram code; it was bundled
in the original PR and is lifted out here as its own reviewable unit.

What's in this PR
- docs/zh/visual-design-paradigm.md: the eight-dimension design system (type
  ladder, weight contrast, spacing, color discipline, ...) + section 13
  feedback-evolution mechanism that seeds the exemplars library.
- templates/exemplars/: the gold-page library (README + first keynote-cover
  exemplar) that section 13 produces.

Review fixes
- F17: a section 4.4 code example used font-size="20", a value the doc's own
  section 2.1 explicitly bans; bumped to 22 (the subtitle tier on the type
  ladder).
- F18: the header cross-reference linked to "the original
  visual-design-paradigm.md", i.e. itself; removed the self-referencing link.

Scope: base main, independent. Docs + one exemplar SVG; no assets / decks.

---
1. #156 core


hugohe3:
@shenqili Thank you for this — the design knowledge is genuinely strong: the Major-Third type ladder, navy-not-black shadows, skip-a-weight contrast, optical-center composition, 70-20-8-2 color budgeting are all real, well-grounded principles, and you clearly understand visual design deeply.

I'm going to hold off on merging this, and I want to be transparent that it's about process and risk, not the quality of your thinking.

This project now has a large number of stars and forks, and at that scale any change can carry outsized, hard-to-reverse consequences. That's especially true of the prompt layer: a substantial rewrite there can shift the behavior of every deck the tool generates, in ways that are hard to predict from the diff alone. So large changes to the prompts and reference docs are something I'm very cautious about taking directly as a PR.

What I'd suggest instead: let's start with an **issue scoped to one specific item** — one section, one idea — and discuss the direction there until we're aligned. Then you open a focused PR for just that piece, and I can test the behavior before and after the change to see its real effect. Landing a 1000-line paradigm in a single step skips that step, and I can't responsibly verify it.

One thing I'll say upfront about the numbers: my consistent position is that this kind of guidance should be offered as a design-spec **reference, not a hard constraint**. The model is generative — that's the whole point of the tool. If we pin it down with fixed quotas and pass/fail thresholds ("use only these values", a 0–10 rubric, "redo under 60"), we're effectively just writing code and giving up the generative capability we rely on. So any version of this I'd want to read as guidance the model can draw on, not rules it must obey.

The design thinking here is excellent — I'd just like to bring it in piece by piece, discussion first. Thanks again for engaging so deeply with the visual layer 🙏

hugohe3:
I took another look at this PR—there is a lot of great content here. I'm still figuring out the best way to integrate it with our existing workflow, so I’d like to take some time (at least over the weekend) to review it thoroughly before making a decision.

Thank you for the hard work; I appreciate your patience. Regarding the native diagram features, let's open an issue to discuss that direction first. I will follow up with more feedback once I've finished reading through everything.

shenqili:
Thanks again for this review — the "reference, not constraint" point landed, and I agree a 1000-line hard-rule block shouldn't go in as one piece.

So I've reframed it instead of defending it: the direction is in #163 (visual guidance as soft reference + personal feedback memory, not fixed quotas), and the first concrete slice is #164 — the type-scale section rewritten from its hard form (banned `font-size` values; a "0–10, redo under 60" rubric) into a soft reference + reflection you can read before→after.

The plan for this doc: rather than land it whole, decompose it section by section into that soft-reference form (#164 is the template) — shadow discipline, color budgeting, spacing, optical centering, and so on — each as its own small, testable PR, so you can judge each piece's real effect instead of a big-bang.

Happy to close this PR in favor of those slices, or keep it open as a reference while we land them — whatever's easiest for your review. No rush; let's settle the form in #163 / #164. 🙏

hugohe3:

PR #158, opened 2026-06-10
feat(native-diagram): recolor / text-fill / font transforms

shenqili:
Layer the three data-* transforms onto the core splice path: the converter
now recolors, text-fills, and re-fonts a spliced component to match the
deck's spec_lock contract.

What's in this PR
- native_diagram_resolver.py: _apply_recolor (remap base srgbClr hexes from
  data-recolor), _apply_text_fills (substitute <a:t> runs by slot id from
  data-text; slot id = document order, matching meta.text_slots), _apply_font
  (force the deck typeface onto every run from data-font).
- references/native-diagrams.md section 3: the data-recolor / data-text /
  data-font contract.

Blocker fixed (your item 1)
data-font is sourced from spec_lock.typography.font_family, which is a CSS
stack (e.g. "Microsoft YaHei", Arial, sans-serif). DrawingML's typeface
attribute accepts only ONE font name, so writing the stack verbatim produced
a broken PPTX. _first_family() now reduces the stack to its first concrete
family (stripping quotes + generic fallbacks) before it reaches
<a:latin>/<a:ea>/<a:cs>. The doc field path is also corrected
(fonts.font_family -> typography.font_family).

Related fixes (same string-into-XML family)
- recolor runs in a SINGLE pass; a per-pair loop corrupted colors when a
  destination hex equalled another source hex.
- data-text ignores non-object JSON instead of raising AttributeError.
- the font-child strip handles paired <a:latin>...</a:latin>, not only the
  self-closing form.
- data-recolor validates each hex pair, skipping malformed ones with a warning.

Scope: stacked on feat/native-diagram-core. Net-new = the resolver transform
block + native-diagrams.md section 3.
Verified: inline asserts against the real functions (CSS-stack -> single
typeface; single-pass recolor on a cascading map; non-dict data-text no-op;
paired <a:latin> stripped).

---
Stacked on #156 -- please review/merge after it (cross-fork base is `main`, so the diff is cumulative until the parent merges).

---
1. #156 core

### DIFF
--- a/skills/ppt-master/references/native-diagrams.md
+++ b/skills/ppt-master/references/native-diagrams.md
@@ -1,0 +1,117 @@
+# Native Diagrams — Selection & Placement Reference
+
+> Lazy-loaded reference. Load **only** when `spec_lock.md page_diagrams` has at
+> least one entry (Strategist matched a page to a native diagram). If that section
+> is empty/absent, this file is never read. Peer to [`image-generator.md`](./image-generator.md)
+> and the chart catalog — same "scan an index, match by content shape" pattern.
+
+A **native diagram** is a pre-designed, fully-editable DrawingML figure lifted
+verbatim from a real deck and stored in `templates/native_diagrams/`. Unlike a
+`templates/charts/` SVG (which Executor *redraws* from scratch using the template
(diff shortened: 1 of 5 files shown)


hugohe3:

hugohe3:

PR #159, opened 2026-06-10
feat(native-diagram): extract / inject / compose engine + CLIs

shenqili:
Component-producing tooling on the core component format: lift a slide into a
library component, inject a component into a deck, compose components, plus
their CLIs.

What's in this PR
- native_diagrams/extract.py: lift a slide's top-level shapes into a
  self-contained, theme-flattened component (shapes.xml.gz + meta.json).
- native_diagrams/inject.py: append a component's shapes into a target deck,
  remapping media rIds.
- native_diagrams/compose.py: combine several components into one.
- extract_diagram.py / inject_diagram.py: the CLIs.
- native_diagrams/__init__.py: package exports.

Blocker fixed (your item 2)
Charts reference their part via <c:chart r:id=...> (attribute r:id), but
detection only checked embed/link, so charts_unsupported was ALWAYS empty
(charts never recorded), AND the chart graphicFrame was lifted (it is in
SHAPE_TAGS) then had its r:id stripped by strip_foreign_rels, shipping an
empty/dangling frame. Both halves of the documented "charts not lifted"
behavior were violated. A chart frame is now detected by its r:id->chart
relationship, excluded from the lifted set, and its rId recorded in
charts_unsupported with a logged note.

Related fixes
- extract: media scanned over lifted shapes only; external-linked images
  (TargetMode=External) skipped so resolve_target + z.read cannot crash on a
  URL target.
- compose: the <a:diagram> wrapper declares the r: namespace, so media refs do
  not serialize as ns0:embed.
- inject: fall back to the last slide layout when a target deck has no 7th
  ("Blank") layout.

Scope: stacked on feat/native-diagram-core. This is the per-component half of
your extract/inject/scan/gallery bucket, split from the scan/gallery half for
tighter scope. build_diagram_index.py is pack-specific curation and travels
with the asset bundle, not here.
Verified: py_compile + package import + a synthetic chart-detection test.

---
Stacked on #156 -- please review/merge after it (cross-fork base is `main`, so the diff is cumulative until the parent merges).

---
1. #156 core

### DIFF
--- a/skills/ppt-master/references/native-diagrams.md
+++ b/skills/ppt-master/references/native-diagrams.md
@@ -1,0 +1,117 @@
+# Native Diagrams — Selection & Placement Reference
+
+> Lazy-loaded reference. Load **only** when `spec_lock.md page_diagrams` has at
+> least one entry (Strategist matched a page to a native diagram). If that section
+> is empty/absent, this file is never read. Peer to [`image-generator.md`](./image-generator.md)
+> and the chart catalog — same "scan an index, match by content shape" pattern.
+
+A **native diagram** is a pre-designed, fully-editable DrawingML figure lifted
+verbatim from a real deck and stored in `templates/native_diagrams/`. Unlike a
+`templates/charts/` SVG (which Executor *redraws* from scratch using the template
+as reference), a native diagram is **spliced in as-is** via a placeholder and only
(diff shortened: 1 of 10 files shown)


hugohe3:

hugohe3:

PR #160, opened 2026-06-10
docs(native-diagram): workflow + spec integration

shenqili:
Wire native diagrams into the generation workflow so a structural page can be
routed to a native diagram end-to-end: Strategist per-page Vehicle Decision ->
spec_lock.md page_diagrams -> Executor placeholder -> converter.

What's in this PR
- strategist.md: native diagram as a per-page structural vehicle + selection
  schema.
- executor-base.md: the data-native-diagram placeholder authoring rules.
- spec_lock_reference.md: the page_diagrams lock section.
- design_spec_reference.md: section VII / IX native-diagram rows.
- SKILL.md: lazy-load pointer.
- docs/zh/generation-logic.md: the per-page vehicle decision logic.

Review fixes
- F2 style-gate: reconcile the "deck-style gated / austere decks must not use"
  language with the asset README/index/code (no hard gate, a soft cohesion
  call) across strategist / design_spec / generation-logic.
- F6: the Executor placeholder example now includes data-font (the
  "always set it" attribute), citing the first family of
  spec_lock.typography.font_family.
- F11: drop the explanatory parenthetical from a section heading
  (prompt-style 2/3).
- F12: replace a stray Chinese word in the English strategist.md with
  "premium" (language consistency).

Scope: stacked on feat/native-diagram-text-font (references the full data-*
contract). Docs only.

---
Stacked on #158 -- please review/merge after it (cross-fork base is `main`, so the diff is cumulative until the parent merges).

---
1. #156 core


hugohe3:

hugohe3:

PR #161, opened 2026-06-10
feat(native-diagram): batch pack scan + gallery

shenqili:
Library-scale tooling on the engine: batch-extract a whole .pptx into the
component library and render an HTML gallery.

What's in this PR
- scan_pack.py: batch-lift every slide of a source deck into the library and
  seed/update diagrams_index.json.
- make_gallery.py: render the index + per-component thumbnails to gallery.html.

Blocker fixed (your item 3)
scan_pack.py read the index raw and wrote it back FLAT ({key: ...}), which
corrupts a curated { meta, diagrams } index (the new format that
build_diagram_index.py and make_gallery.py already use). It now reads
raw.get("diagrams", raw), preserves meta, writes the { meta, diagrams } shape,
and upgrades a legacy flat file in place.

Scope: stacked on feat/native-diagram-engine (scan_pack imports
extract_diagram). The scan/gallery half of your tooling bucket.
Verified: py_compile + a synthetic round-trip test (preserves meta; upgrades
legacy flat; never writes flat).

---
Stacked on #159 -- please review/merge after it (cross-fork base is `main`, so the diff is cumulative until the parent merges).

---
1. #156 core

### DIFF
--- a/skills/ppt-master/references/native-diagrams.md
+++ b/skills/ppt-master/references/native-diagrams.md
@@ -1,0 +1,117 @@
+# Native Diagrams — Selection & Placement Reference
+
+> Lazy-loaded reference. Load **only** when `spec_lock.md page_diagrams` has at
+> least one entry (Strategist matched a page to a native diagram). If that section
+> is empty/absent, this file is never read. Peer to [`image-generator.md`](./image-generator.md)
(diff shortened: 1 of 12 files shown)


hugohe3:

hugohe3:

PR #185, opened 2026-06-16
Update windows-installation.md

BonaSkeleton:
easyer to download requirements.txt

### DIFF
--- a/docs/windows-installation.md
+++ b/docs/windows-installation.md
@@ -51,7 +51,7 @@ pip install -r requirements.txt
 ```
 
 > If `pip` is not recognized, try `python -m pip install -r requirements.txt`.
-
+> or try  `pip install -r skills/ppt-master/requirements.txt`.   
 Wait for it to finish. You should see `Successfully installed ...` at the end.
 
 ---


hugohe3:
Thanks for taking the time to send this, but I'm going to decline this one.

Two reasons:

1. **It's redundant.** The root `requirements.txt` is a thin wrapper whose only content is `-r skills/ppt-master/requirements.txt` (the full dependency list lives inside the skill on purpose). Step 3 already does `cd ppt-master` and then `pip install -r requirements.txt`, which from the repo root installs the complete set. The suggested alternative path resolves to the exact same install, just via a longer route.

2. **It would break the rendering.** The new line replaces the blank line between the tip blockquote and `Wait for it to finish.`, so that sentence becomes a lazy continuation of the blockquote and gets pulled inside the quote (there are also trailing spaces).

More generally: for one-line doc/usage corrections like this, please open an issue rather than a PR — it's faster for me to apply directly and keeps the history clean. See CONTRIBUTING.md ("Tiny fixes" / "Please don't send PRs for…"). Appreciate the intent though!

PR #200, opened 2026-06-25
fix: account for letter-spacing and serif width in text box sizing (#199)

### LINKED ISSUE #199: Native PPTX text wraps/overflows in LibreOffice-rendered PDFs: converter ignores letter-spacing & underestimates serif glyph width
## Summary

When a native (DrawingML) PPTX is rendered to PDF by **LibreOffice** (`soffice --headless --convert-to pdf`), single-line text boxes wrap their last word/letter onto a second line — e.g. an all-caps title `PSYOPTIMAL` renders as `PSYOPTIMA` / `L`, and letter-spaced eyebrow/kicker labels (`MARKET OPPORTUNITY`, `PROJECTED BY 2030`) wrap and collide with adjacent elements. The same PPTX looks correct in PowerPoint/Keynote.

## Root cause

In `scripts/svg_to_pptx/drawingml_elements.py::convert_text`, single-line text boxes are emitted with `wrap="none"` + `<a:spAutoFit/>`. PowerPoint shrink-wraps the frame, so it's fine there — but **LibreOffice ignores `wrap="none"`** and wraps text to the box width `cx`. `cx` comes from `estimate_text_width()` (`scripts/svg_to_pptx/drawingml_utils.py`), which:

1. **ignores `letter-spacing`** entirely — eyebrow labels use `letter-spacing` 1–3px, adding 15–50px of real width that isn't counted; and
2. uses a **0.55em average** char width that significantly under-sizes wide **Georgia/serif caps** (and a bold `<tspan>` inside a normal-weight `<text>` is measured as normal).

The box ends up narrower than the rendered glyphs → LibreOffice wraps.

## Repro

Author an SVG `<text>` with `font-family="Georgia"` + an all-caps string, or any `letter-spacing`d label sized near its content width; convert with `--only native`; render the PPTX via LibreOffice → PDF. The last token wraps.

## Fix (verified)

In `convert_text`, add the tracking to every measured width and apply a serif-aware single-line margin. This is safe because runs stay anchored within the box via `algn` (l/ctr/r), so a wider box never shifts visible text, and real PowerPoint still `spAutoFit`s — only the LibreOffice render gains headroom:

```python
# parse letter-spacing (px) once
ls_px = float(letter_spacing) * (ctx.scale_x or 1.0)  # 0.0 if unset
is_serif = any(f in font_family.lower() for f in ("georgia", "times", "serif"))
single_line_fudge = 1.40 if is_serif else 1.18

# single-line width
text_width = (estimate_text_width(full_text, font_size, font_weight) * single_line_fudge
              + ls_px * max(len(full_text) - 1, 0))
# and add ls_px * (len(line)-1) to each visual_line_widths entry
```

Happy to open a PR with this if useful.

JFJLL:
## Fix: LibreOffice wraps single-line text due to underestimated box width

Fixes #199

### Problem

When converting SVG `<text>` to DrawingML, `convert_text()` underestimates the text box width in two ways:

1. **Letter-spacing ignored** — `letter-spacing` (tracking) is parsed and applied to the DrawingML `spc` attribute, but NOT factored into the box width calculation. Eyebrow/kicker labels with 1-3px tracking add 15-50px of real width that the box doesn't account for.

2. **Serif fonts underestimated** — `estimate_text_width()` uses a 0.55em average character width, which significantly undersizes wide Georgia/serif capitals (e.g. `PSYOPTIMAL` renders as `PSYOPTIMA` / `L`).

PowerPoint ignores the issue because `<a:spAutoFit/>` shrink-wraps the frame. But **LibreOffice ignores `wrap="none"`** and wraps to the box `cx`, exposing the underestimation.

### Fix

**`drawingml_utils.py` — `estimate_text_width()`**
- Added `font_family` parameter: serif fonts (Georgia, Times, etc.) get a 1.25x multiplier on average glyph width
- Added `letter_spacing` parameter: adds `ls * (len(text) - 1)` px to account for inter-character gaps
- Both parameters are optional with zero-effect defaults, so all existing callers are unaffected

**`drawingml_elements.py` — `convert_text()`**
- Moved `letter-spacing` parsing before the width calculation so `ls_px` is available
- Passes `font_family` and `ls_px` to `estimate_text_width()`

### Tests

11 pytest cases in `tests/test_text_width.py`:
- Basic width estimation (ASCII, bold, CJK)
- Letter-spacing: adds width, zero has no effect, single-char unaffected
- Serif awareness: Georgia/Times wider, Arial unaffected
- Combined: Georgia caps + tracking, eyebrow label scenario

### DIFF
--- a/skills/ppt-master/scripts/svg_to_pptx/drawingml_elements.py
+++ b/skills/ppt-master/scripts/svg_to_pptx/drawingml_elements.py
@@ -800,8 +800,23 @@ def convert_text(elem: ET.Element, ctx: ConvertContext) -> ShapeResult | None:
     if not full_text.strip():
         return None
 
+    # Letter spacing — parse once, used for both box width and DrawingML spc attr
+    spc_attr = ''
+    letter_spacing = _get_attr(elem, 'letter-spacing', ctx)
+    ls_px = 0.0
+    if letter_spacing:
+        try:
+            ls_val = float(letter_spacing)
+            ls_px = ls_val * (ctx.scale_x or 1.0)
+            spc_attr = f' spc="{int(ls_val * 100)}"'
+        except ValueError:
+            pass
+
     # Estimate text dimensions
-    text_width = estimate_text_width(full_text, font_size, font_weight) * 1.15
+    text_width = estimate_text_width(
+        full_text, font_size, font_weight,
+        font_family=font_family_str, letter_spacing=ls_px,
+    ) * 1.15
     text_height = font_size * 1.5
     padding = font_size * 0.1
 
@@ -816,16 +831,6 @@ def convert_text(elem: ET.Element, ctx: ConvertContext) -> ShapeResult | None:
     box_y = y - font_size * 0.85
     box_w = text_width + padding * 2
     box_h = text_height + padding
-
-    # Letter spacing
-    spc_attr = ''
-    letter_spacing = _get_attr(elem, 'letter-spacing', ctx)
-    if letter_spacing:
-        try:
-            spc_val = float(letter_spacing) * 100
-            spc_attr = f' spc="{int(spc_val)}"'
-        except ValueError:
-            pass
 
     # Text rotation. SVG's rotate(angle [cx cy]) rotates around (cx, cy), but
     # DrawingML's <a:xfrm rot="..."> rotates the shape around its own center.
--- a/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
+++ b/skills/ppt-master/scripts/svg_to_pptx/drawingml_utils.py
@@ -281,8 +281,26 @@ def is_cjk_char(ch: str) -> bool:
             0x20000 <= cp <= 0x2A6DF)
 
 
-def estimate_text_width(text: str, font_size: float, font_weight: str = '400') -> float:
-    """Estimate text width in SVG pixels."""
+SERIF_FONTS = ('georgia', 'times', 'palatino', 'garamond', 'serif')
+
+
+def estimate_text_width(
+    text: str,
+    font_size: float,
+    font_weight: str = '400',
+    font_family: str = '',
+    letter_spacing: float = 0.0,
+) -> float:
+    """Estimate text width in SVG pixels.
+
+    Accounts for letter-spacing (tracking) and applies a serif-aware
+    multiplier so that wide-caps fonts like Georgia produce a text box
+    large enough for LibreOffice (which ignores ``wrap="none"``).
+    """
+    # Serif fonts have wider average glyphs than the 0.55em baseline.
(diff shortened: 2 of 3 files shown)


hugohe3 (changes requested):
Thanks for working on this. The issue is still valid on the current main branch, and the overall direction of accounting for tracking and serif/wide text in the box width estimate makes sense.

That said, this PR needs changes before it can be merged:

1. Please rebase this PR on the current main branch. GitHub currently marks it as conflicting, and the text export code has changed around paragraph merging.

2. Please preserve the current estimate_text_width() behavior from main. In particular, main already treats % as a wide character and handles digits separately so 1 is not underestimated. This PR appears to be based on an older version and would drop those fixes.

3. Please apply the width fix to both current text paths. main now measures visual_line_widths for paragraph mode and full_text for non-paragraph mode. Letter-spacing and serif-aware sizing need to affect both, otherwise merged paragraphs can still be underestimated.

4. Please avoid substring matching for serif detection. Checking whether serif appears in the raw font-family string can misclassify sans-serif as serif and widen normal text boxes. Prefer parsing/resolving the font family and checking against the existing serif font set.

5. Please remove tests/test_text_width.py. This repository intentionally does not ship automated tests; docs/rules/code-style.md forbids tests/ directories and test_*.py files. Please use a smoke command and manual verification notes in the PR description instead.

Once these are addressed, I am happy to review the updated patch again.

hugohe3:
Thanks for the careful write-up on #199 and for sending this — the root-cause analysis is correct: the single-line box width did underestimate both `letter-spacing` (tracking) and wide serif caps, which is what made LibreOffice (ignoring `wrap="none"`) wrap the last token.

After a close review, though, we're going to close this PR, because `main` already handles the same problem with a more complete implementation:

- Width estimation and the DrawingML `spc` output are now done at the **run level** in `drawingml_elements.py` (`_parse_letter_spacing_px`, `_letter_spacing_to_drawingml_spc`, `_estimate_text_runs_width`, `_SERIF_WIDTH_FAMILIES`), plus a follow-up `avoid paragraph textbox width headroom` pass. Working per-run rather than per-string lets it measure a bold `<tspan>` inside a normal-weight `<text>`, parse `letter-spacing` units (px/pt/em), and stay correct under paragraph-merge — cases this PR's whole-string approach in `estimate_text_width()` can't cover.
- As a result the PR now conflicts with `main` (CONFLICTING / DIRTY).
- This repo also deliberately ships no automated tests — `tests/` directories and `test_*.py` are disallowed by `docs/rules/code-style.md` §11 — so `tests/test_text_width.py` can't be merged.

The diagnosis was spot-on and matched our own; it just landed in `main` first via a different (run-level) structure. Really appreciate the issue and the PR — thank you!

PR #202, opened 2026-06-27
feat(image-backends): add Agnes Image 2.1 Flash backend

xiaozhangpgg:
- Added `backend_agnes.py`: Agnes Image 2.1 Flash text-to-image backend
  - Registered Agnes as EXPERIMENTAL backend in `image_gen.py`
  - Endpoint: POST https://apihub.agnes-ai.com/v1/images/generations
  - Auth: Bearer Token via AGNES_API_KEY env var
  - Supports 5 aspect ratios × 4 resolution tiers (512px/1K/2K/4K)

### DIFF
--- a/skills/ppt-master/scripts/image_backends/backend_agnes.py
+++ b/skills/ppt-master/scripts/image_backends/backend_agnes.py
@@ -1,0 +1,176 @@
+#!/usr/bin/env python3
+"""
+Agnes Image 2.1 Flash generation backend.
+
+Configuration keys:
+  AGNES_API_KEY   (required)
+  AGNES_BASE_URL  (optional)
+  AGNES_MODEL     (optional)
+"""
+
+import sys
+from pathlib import Path
+
+_SCRIPTS_DIR = Path(__file__).resolve().parents[1]
+if str(_SCRIPTS_DIR) not in sys.path:
+    sys.path.insert(0, str(_SCRIPTS_DIR))
+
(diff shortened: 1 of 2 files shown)


hugohe3:
Thanks for the contribution. I'm closing this PR because Agnes Image 2.1 Flash is already supported through the existing OpenAI-compatible image backend.

Use this configuration instead:

```env
IMAGE_BACKEND=openai
OPENAI_API_KEY=your-agnes-key
OPENAI_MODEL=agnes-image-2.1-flash
OPENAI_BASE_URL=https://apihub.agnes-ai.com/v1
OPENAI_SIZE_PRESET=gpt-image-2
OPENAI_RESPONSE_FORMAT=omit
OPENAI_QUALITY=omit
```

Adding a separate `agnes` backend duplicates the existing compatibility path and goes against the current backend design documented in `image-generator.md`, which keeps OpenAI-compatible providers under `IMAGE_BACKEND=openai` with provider-specific compatibility knobs.

There is also a concrete URL handling issue in this PR: if users set `AGNES_BASE_URL=https://apihub.agnes-ai.com/v1`, the new backend appends `/v1/images/generations` again and would request `.../v1/v1/images/generations`.

If Agnes later requires behavior that cannot be represented by the OpenAI-compatible backend, we can revisit this with a focused change that includes the URL resolver, docs, env examples, and minimal validation coverage.

PR #203, opened 2026-06-29
Publish PPT Master skill under AGPL-3.0

Runa798:
## Summary
- Switch project license metadata and LICENSE text from MIT to GNU Affero General Public License v3.0
- Keep the Claude/Codex skill packaging metadata aligned with AGPL-3.0
- Add presentation quality gates and wire them into Strategist, Executor, design spec, and spec lock templates
- Ignore local .venv directories

## Checks
- python3 -m json.tool .claude-plugin/marketplace.json
- git diff --check
- secret-pattern scan for common PAT/API-key formats


hugohe3:
Thanks for the PR, but neither change is something I'll merge.

**License**: the project has been MIT from the start, and that's a deliberate choice — it won't change. The open-source license is a foundational setting of the project; there's no reason to switch it halfway through. Closing that part outright.

**Quality gates**: this project deliberately avoids fixed numeric quotas (`max_cards` / `max_bullets` / `max_table_rows` and similar hard caps) to constrain generation. Layout density is governed by narrative rhythm (the existing `page_rhythm`) and one primary focus per page; storyline and one-message-per-page are already covered by the `design_spec.md §IX` content outline and the Eight Confirmations flow. This gate layer overlaps with the existing mechanisms and runs against the project's direction, so I won't adopt it.

If you run into specific generation problems down the line, feel free to open an issue with a sample and we can look at it together.

PR #207, opened 2026-07-01
feat: docx using markitdown

alvenchen:
Changes:
  - doc_to_md.py: Adds _convert_docx_with_markitdown as the preferred path; falls back to the existing mammoth path
  (_convert_docx_with_mammoth) when markitdown is unavailable or raises an error
  - Since markitdown uses mammoth internally for images and emits them as truncated data: URIs (binary never reaches the markdown),
  adds _extract_docx_images_to_media to extract every word/media/* part directly into <input>_files/, and
  _rewrite_markitdown_image_refs to rewrite placeholders in document order with relative paths — preserving the same
  image_manifest.json output shape as the rest of the source_to_md tools
  - markitdown's own pre_process_docx preserves OMML/Office Math as $...$ / $$...$$ LaTeX, so the math-injection pass from the mammoth
  path is skipped
  - requirements.txt: Adds markitdown[docx]>=0.1.0; keeps mammoth as fallback dependency
  - conversion.md: Updates dependency notes and install commands to reflect the new primary path and fallback behavior

### DIFF
--- a/skills/ppt-master/requirements.txt
+++ b/skills/ppt-master/requirements.txt
@@ -46,10 +46,11 @@ PyMuPDF>=1.23.0
 # ─────────────────────────────────────────────────────────────
 # Native (pure-Python) paths — no external binary required:
 # 原生路径（纯 Python，无需外部程序）：
-#   .docx   → mammoth
+#   .docx   → markitdown (preferred), mammoth fallback
 #   .html   → markdownify (+ beautifulsoup4, listed below)
 #   .epub   → ebooklib (+ markdownify)
 #   .ipynb  → nbconvert
+markitdown[docx]>=0.1.0
 mammoth>=1.6.0
 markdownify>=0.11.6
 ebooklib>=0.18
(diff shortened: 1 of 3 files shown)


hugohe3:
Thanks for this PR, and for surfacing the gap.

I tested markitdown 0.1.6 against the current mammoth path across headings, lists, tables, images, and OMML equations. I don't think we should merge this as-is, but it did point at something worth fixing on our side.

The blocker is image handling. Extracting `word/media/*` by filename order and rewriting the `data:` placeholders positionally breaks whenever media-part order doesn't match document order, or an image is referenced more than once. In my test case (4 image references, 3 media parts because Word de-dupes a reused image), the reused image resolved to the wrong file, and the extra reference was left as a full base64 `data:` URI in the Markdown.

The hand-built manifest also drops `source_sha256`, Word display sizing, and the accurate `usage_count`/occurrences that `analyze_images.py` relies on. The existing path avoids this by mapping images via document draw order + sha256.

The concrete markitdown gains I could confirm are:

- **Tables**: mammoth flattens them, while markitdown emits pipe tables.
- **Inline text**: mammoth backslash-escapes punctuation and can break URLs, e.g. `https://example\.com`, while markitdown output is cleaner.

Everything else is broadly on par: headings and lists match, and OMML -> LaTeX is already handled in our current path via `_docx_inject_math_latex`.

We've now strengthened those two areas directly on the existing mammoth route (text-only tables to pipe Markdown, plus trimming the unnecessary escaping), without swapping the DOCX engine or adding the markitdown dependency. Really appreciate you flagging this — that's the useful outcome from the PR.

PR #217, opened 2026-07-03
feat(i18n): add Japanese as a third UI language (Confirm UI + SVG editor)

kevindesuyo:
This adds Japanese as a third interface language next to 中文 / English, covering both local web UIs.

**Confirm UI**
- Full ja message dictionary (83 keys, same key set as zh/en) and `label_ja` / `desc_ja` / `group_ja` for every catalogs.json entry.
- `localized()` accepts `_ja` fields from AI-authored recommendations.json (`name_ja` / `note_ja` / `tag_ja`), following the same suffix pattern as `_zh` / `_en`.

**SVG editor**
- Full ja dictionary (74 keys).

**Language switcher**
The 中 ⇄ EN cycle button becomes a small custom dropdown (中文 / English / 日本語) — a two-way toggle stops working once there are three languages. It is keyboard-operable (arrows / Enter / Esc / Home / End) with listbox ARIA roles, and while the menu is open its key events don't reach page-level shortcuts (in the SVG editor, arrow keys would otherwise nudge the selected element).

**Docs**
The trilingual contract is reflected where the schema is owned: SKILL.md step 4, `scripts/docs/confirm_ui.md`, `references/strategist.md` (spectrum tags), the `workflows/beautify-pptx.md` sample, and `workflows/live-preview.md`.

**Nothing changes for existing users**
- Locale detection: zh-* still gets 中文 and every other locale still gets English; only ja-* now gets Japanese. `ppt_lang` in localStorage and recommendations.json `lang` behave as before.
- Label fallback stays language-relative: zh/en pages keep their zh↔en fallback and only try `_ja` last; a ja page falls back ja → en → zh.
- No schema changes: result.json output and existing recommendations.json inputs (multilingual or legacy `name` / `note`) load unchanged.

Tested locally by driving the Confirm UI in a browser in all three languages with a trilingual recommendations.json; both app bundles pass `node --check` and catalogs.json stays valid JSON.


hugohe3:
Merged, thanks.

This was reasonable and well-scoped: Japanese is covered across both local UIs, existing zh/en behavior stays compatible, the catalog coverage is complete, and replacing the two-way toggle with a language dropdown is the right shape now that there are three languages.

For future language additions, let's keep the same bar: full UI dictionaries, complete catalog labels/descriptions, explicit fallback behavior, browser smoke coverage, and docs sync. I noticed one non-blocking wording leftover (`confirm_ui/server.py` still says “bilingual labels” in an internal comment); that can be handled as follow-up cleanup rather than blocking this merge.

kevindesuyo:
Thanks for the quick merge, and good catch on the server.py docstring  I've opened #219 with that wording fix. While at it I checked the repo's other "bilingual" mentions; the two remaining ones are about mixed CJK/Latin deck content and template naming rather than the UI language count, so I left them alone.

Noted on the bar for future language additions  happy to follow the same checklist if I bring another one.

Really glad to have found this project. ppt-master has quietly become part of my daily deck workflow, and getting to use it in Japanese now makes it even better. Thanks for building and maintaining it.

hugohe3:
Thanks Kevin, I appreciate the quick follow-up. #219 is merged now as well.

The sweep of the remaining “bilingual” references was helpful, and leaving those two alone makes sense given their context. Thanks also for being explicit about following the same completeness bar for any future language additions.

Really glad to hear ppt-master has become part of your daily deck workflow. Japanese support is a valuable addition, and this PR kept the implementation focused while preserving the existing zh/en behavior. Thanks for contributing it.

PR #221, opened 2026-07-03
fix: deterministically backfill root <svg> width/height from viewBox

TubeLiu:
## Problem

A root `<svg>` carrying only `viewBox` (no `width`/`height`) is valid, scalable SVG — which is exactly why generated slides often come out that way. But it trips PPT preview/export dimension detection:

- the **live-preview server** fires its `SVG is missing width/height attributes` banner on every such slide;
- the **quality checker** silently passes it — worse, its `width="(\d+)"` search scans the whole document and matches a child `<rect>`, so a dimensionless root **false-passes**.

Relying on every model to remember the attributes is fragile. `viewBox` already carries the exact canvas size, so restoring `width`/`height` from it is **lossless and model-independent**.

## Change (defense in depth)

- **New `svg_finalize/normalize_dimensions.py`** — viewBox → width/height backfill, with an **ET adapter** (preview serve path) and a **string adapter** (finalize) sharing one viewBox parser so they can't drift. Only integer `0 0 W H` viewBoxes are honored; an author-supplied `width`/`height` is never overridden.
- **`finalize_svg.py`** — new always-on `backfill-dims` step (runs last, over `svg_final`).
- **`svg_editor/server.py`** — backfill the root in `get_slide` before serving (in memory only; `svg_output` on disk stays raw), suppressing the banner regardless of which model wrote the SVG.
- **`svg_quality_checker._check_dimensions`** — scope the check to the root `<svg>` tag (fixes the child-`<rect>` false-pass) and **warn** instead of silently passing when the root omits `width`/`height`. It stays a warning, not an error, because the pipeline already auto-backfills — erroring would false-fail decks that render correctly.
- **`shared-standards.md §4`** — state the root MUST carry explicit `width`/`height` equal to `viewBox`, and note the pipeline auto-backfills as a safety net.

## Tests

Adds `scripts/tests/test_svg_dimensions.py` (12 tests, pure stdlib): both backfill adapters, child-element isolation, and the quality-gate behavior (warns on missing / no longer false-passes on child width / consistency / mismatch).

```
python -m pytest skills/ppt-master/scripts/tests/test_svg_dimensions.py
# 12 passed
```

Also verified end-to-end: `finalize_svg.py --only backfill-dims` restores `width="1280" height="720"` on `svg_final` while leaving the raw `svg_output` untouched.

### DIFF
--- a/skills/ppt-master/references/shared-standards.md
+++ b/skills/ppt-master/references/shared-standards.md
@@ -169,7 +169,7 @@ One offending character invalidates the file and aborts export. Numeric refs (`&
 
 ## 4. Basic SVG Rules
 
-- **viewBox** must match the canvas dimensions (`width`/`height` must match `viewBox`)
+- **Root `<svg>` MUST carry explicit `width` and `height` attributes**, and both MUST equal the `viewBox` size — `viewBox="0 0 W H"` → `width="W" height="H"`. A root with only `viewBox` renders fine in browsers but trips PPT preview/export dimension detection (the live-preview banner "SVG missing width/height" fires). The finalize + live-preview pipeline auto-backfills missing `width`/`height` from `viewBox` as a safety net — do not rely on it; write them explicitly.
 - **Background**: Use `<rect>` to define the page background color
 - **`<tspan>`** has two purposes: (1) manual line breaks (use `dy` or explicit `y`); (2) inline run formatting on the same line (color/weight/size). `<foreignObject>` is FORBIDDEN. See "Single logical line" rule below.
 - **Fonts**: every `font-family` stack MUST end with a pre-installed family (Microsoft YaHei / SimSun / Arial / Times New Roman / Consolas …); `@font-face` is FORBIDDEN. Full rule: [`strategist.md §g`](strategist.md).
--- a/skills/ppt-master/scripts/finalize_svg.py
+++ b/skills/ppt-master/scripts/finalize_svg.py
@@ -31,6 +31,7 @@ Processing options:
                     merged step, so existing --only invocations keep working.
     flatten-text  - Convert <tspan> to independent <text> (for special renderers)
     fix-rounded   - Convert <rect rx="..."/> to <path> (for PPT shape conversion)
+    backfill-dims - Restore root <svg> width/height from viewBox when missing
 """
 
 import os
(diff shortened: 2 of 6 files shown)


hugohe3:
Thanks for the detailed PR. After a full review I'm closing it — explanation below, including the complete history of how this attribute has been handled in this project.

**Timeline of dimension handling**

- **2025-12 (repo start) – 2026-02**: export sizing came straight from the canvas-format config (`CANVAS_FORMATS[...]['dimensions']`); `viewBox` was only used to *detect* the format — root `width`/`height` never participated.
- **2026-02-03** (e42c9ce): PPTX dimensions became dynamically derived from the SVG `viewBox`, including custom canvases.
- **2026-05-19** (#115): a user reported a blank live preview — root cause was a dimensionless root plus the preview CSS (`width/height:auto`) collapsing the box. This is the **only** place missing `width`/`height` ever had a real effect.
- **2026-05-22** (#127): the fix chosen at the time was a red warning banner, deliberately not a backfill ("No backfill — upstream spec stays authoritative").
- **2026-06-18** (95b3b63): on conflict between `viewBox` and the project's recorded canvas format, `viewBox` wins — the chain of truth kept converging on `viewBox`.
- **2026-07-03** (the audit this PR triggered): I verified every consumer — DrawingML conversion (`_root_viewport_size` prefers `viewBox`), export sizing (`pptx_dimensions.py`), both PNG snapshot renderers (cairosvg and svglib render viewBox-only roots correctly; tested), visual-review screenshots (hardcoded CSS sizing). Conclusion: **apart from the preview CSS, nothing consumes root `width`/`height`.**

**So the description's "trips PPT preview/export dimension detection" doesn't hold for export** — no version of the export path has ever read these attributes; a dimensionless root does not affect the final output.

**The fix also goes in a different direction.** The audit pushed the rule to its root: rather than rewriting files at finalize/serve time to satisfy an attribute nobody consumes, we removed the requirement itself. Landed on main (76326a1): shared-standards.md §4 now makes `viewBox` the single source of truth with root `width`/`height` optional; the preview derives the attributes from `viewBox` at render time (view-layer only — files on disk stay untouched) and the red banner is gone; the quality checker's dimension check is deleted — checks should guard content that downstream actually consumes. Your observation that the old whole-document regex matched a child element was correct, and is resolved by the check's removal.

Human-verified, evidence-backed contributions remain very welcome.

PR #223, opened 2026-07-04
bug fixes:image_gen.py: top level must be a JSON object

astonish921:
…generated `images/image_prompts.json` file by the large model is often an array `[]` instead of an object `{}`, which causes exceptions during processing in `image_gen.py`. Add stricter constraint descriptions.

## What & why

python3 skills/ppt-master/scripts/image_gen.py --render-md projects/ppt169_teen_screen_time_control_ppt169_20260704/images/image_prompts.json || python skills/ppt-master/scripts/image_gen.py --render-md projects/ppt169_teen_screen_time_control_ppt169_20260704/images/image_prompts.json

Error: projects/ppt169_teen_screen_time_control_ppt169_20260704/images/image_prompts.json: top level must be a JSON object, got list

reason: generated `images/image_prompts.json` file by the large model is often an array `[]` instead of an object `{}`, which causes exceptions during processing in `image_gen.py`. 

## Verification
 Add stricter constraint descriptions. test is ok.

## Confirmations

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] I personally reviewed the full diff and verified every claim in this description against this repository's actual code (purely AI-generated PRs without human review are closed unmerged)


hugohe3:
Thanks for the careful write-up — the failure mode is real and your diagnosis is accurate: if `image_prompts.json` is emitted as a top-level array, `load_manifest()` raises `top level must be a JSON object, got list`.

That said, I'm going to **close this PR**, and I want to be transparent about the reasoning, because it reflects a consistent principle in this project rather than a judgment on your work:

**1. This isn't an `image_gen.py` bug.** The manifest contract is `{ "items": [...] }`, and `load_manifest()` rejecting a bare array is *correct* schema validation, not a defect. The script's job is to enforce the contract, not to absorb malformed input.

**2. The shape is already specified.** `image-generator.md §6` already documents `{ "items": [...] }` as the manifest schema. So a top-level array isn't an under-specification we need to patch with more prose — it's an agent failing to follow a spec that already exists. Restating the same rule in four more files won't change what a capable agent already gets right, nor reliably fix what a weaker agent gets wrong.

**3. Downstream doesn't exist to cover for upstream.** Whoever produces the artifact owns its correctness. If we silently coerced a bare array into `{ "items": ... }`, we'd also be silently swallowing the fact that such output usually *also* dropped `deck_style_anchor` / `color_scheme` — the palette-truth fields our cross-page visual consistency depends on. Hiding a contract violation is worse than failing loudly on it. The loud error is the intended behavior.

**4. We don't grow the prompt to chase weaker models.** The prompt is already near its working budget, and this project is built for agents capable of following an already-stated contract — not to be padded indefinitely to accommodate a model that can't. That's a capability gap on the agent side, and the right place to fix it is the agent, not our docs.

If there's anything here we'd consider, it's making the *error message itself* more actionable (e.g. hinting the `{ "items": [...] }` wrapper) — a deterministic, zero-prompt-cost aid for the in-pipeline retry loop of any model. But that's a one-line change at the point of failure, not four documentation restatements.

Thanks again for taking the time, and please don't let this discourage future contributions 🙏

PR #229, opened 2026-07-08
feat(svg_to_pptx): keep only the latest pptx at exports/ root, auto-archive older ones

kevindesuyo:
## What & why

In default-flow mode (no `-o`), every export appends another timestamped pptx to `exports/`, so after a few iterations the actual deliverable has to be picked out of a growing pile of near-identical files.

This PR keeps the `exports/` root clean: after a **successful** export, every pre-existing pptx in the `exports/` root (plus its `.trace.json` sidecar) is moved into `exports/archive/`, leaving the freshest export as the only pptx directly under `exports/`.

Behavior details:

- Runs only after the export succeeds — a failed export never disturbs the root
- Explicit `-o` output paths are user-managed and are never archived
- Same-run siblings stay together in the root (`_svg.pptx` snapshot via `--svg-snapshot`, `_native_charts.pptx` via `--native-objects`)
- Name collisions in `archive/` get a numeric suffix; a locked/unmovable file logs a warning and is skipped instead of aborting the sweep
- `template_fill_pptx`'s latest-export lookup (`exports/*.pptx` non-recursive glob) still resolves correctly — arguably more reliably, since the root now only ever contains the latest export

Implementation is a single module-level helper `_archive_previous_exports()` in `svg_to_pptx/pptx_package/cli.py` plus a two-line call site in `main()`; docs updated to match (SKILL.md Step 7.3 output listing, `scripts/docs/svg-pipeline.md`, project README template in `project_manager.py`).

## Verification

Ran on a minimal test project (single-SVG `svg_output/`), all in default-flow mode unless noted:

1. First export → `exports/proj_<ts>.pptx` created, no archive activity (nothing stale)
2. Second export → previous pptx moved to `exports/archive/`, root contains only the new export
3. `--svg-snapshot` export → both same-run files (`.pptx` + `_svg.pptx`) stay in root, prior export archived
4. `-o exports/manual.pptx` → no archiving performed, root left untouched

## Confirmations

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] I personally reviewed the full diff and verified every claim in this description against this repository's actual code — including that the problem is real here and the capability isn't already provided by an existing path or already fixed on `main` (purely AI-generated PRs without human review are closed unmerged)
- [ ] This PR is code-only, **or** it touches prompt/instruction text (`SKILL.md`, `references/*.md`, `workflows/*.md`, …) and was discussed and agreed in an issue first — link it here: #___

### DIFF
--- a/skills/ppt-master/SKILL.md
+++ b/skills/ppt-master/SKILL.md
@@ -640,6 +640,7 @@ Default raster handling for `svg_final/`: images are embedded at the rendered SV
 python3 ${SKILL_DIR}/scripts/svg_to_pptx.py <project_path>
 # Output (default-flow mode):
 #   exports/<project_name>_<timestamp>.pptx           ← native pptx (canonical output, reads svg_output/)
+#   exports/archive/                                  ← older exports auto-moved here; exports/ root keeps only the latest
 #   backup/<timestamp>/svg_output/                    ← Executor SVG source backup (always written)
 #
 # Add --svg-snapshot to additionally emit the SVG-image preview pptx alongside the native pptx:
--- a/skills/ppt-master/scripts/docs/svg-pipeline.md
+++ b/skills/ppt-master/scripts/docs/svg-pipeline.md
@@ -45,6 +45,7 @@ python3 scripts/svg_to_pptx.py <project_path> --recorded-narration audio
 Behavior:
 - Default output (default-flow mode, no `-o`):
   - `exports/<project_name>_<timestamp>.pptx` — native editable pptx (canonical output)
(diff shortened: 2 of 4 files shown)


kevindesuyo:
Proposal issue as required by CONTRIBUTING.md for the prompt/instruction-text edits: #230. Keeping this PR in draft until the issue is discussed.

hugohe3:
Closing per the decision in #230 — auto-archiving isn't being taken as a behavior change (default or opt-in); the rationale, plus the naming improvement it prompted (`_narrated` suffix, landed in 37427f94), is over there. The diff itself was solid and the description's claims all verified cleanly against the code, which made this an easy review even though the direction didn't fit. Thanks!

Issue #222, opened 2026-07-04
Feature request: image-to-image / image editing support in image_gen.py


765007025:
## Summary

`image_gen.py` and every backend under `scripts/image_backends/` currently only support text-to-image generation — every backend's `generate(prompt: str, ...)` signature takes a text prompt only, and the OpenAI backend calls `/v1/images/generations`. There's no way to pass a reference/source image for image-to-image editing.

## What I verified

Using my own `OPENAI_BASE_URL` + `OPENAI_API_KEY` config (OpenAI-compatible proxy, `gpt-image-2` model — same config PPT Master already uses via `IMAGE_BACKEND=openai`), I called the OpenAI-compatible `/v1/images/edits` endpoint directly (bypassing `image_gen.py`) with a source image + a text instruction to change its shape/color. It returned 200, and pixel-sampling the result confirmed the edit was actually applied to the input image (not just a fresh unrelated text-to-image generation ignoring the input).

So the underlying API surface for image-to-image already exists on at least the OpenAI-compatible path; `image_gen.py` / `backend_openai.py` just doesn't expose it.

## Why this would help

For PPT generation, image-to-image is useful for: iterating on a previously generated image (style tweak, palette change) without a full re-roll, and editing/restyling a user-provided source image to fit the deck's confirmed visual style — both common asks in the Strategist → Image_Generator flow (`§VIII Image Resource List` / `image_prompts.json`).

## Possible direction (open to your call on design)

- Add an optional reference-image path to the single-image CLI and/or an optional field per item in `image_prompts.json` (manifest mode), so existing `Acquire Via: ai` rows could opt into "edit this existing file" instead of "generate from scratch"
- Start with the OpenAI backend (Core tier, endpoint already confirmed working) as the first implementation; other backends that support editing (Gemini, Qwen, Volcengine, Zhipu, etc.) could follow the same per-backend pattern if their APIs support it
- Backends without an edit API would just not support the new parameter (same tiering logic already used for backend capability differences)

Opening this per `CONTRIBUTING.md`'s guidance to discuss direction before a PR, since this touches `image_gen.py`'s interface and potentially multiple backend files. Happy to submit a focused PR (starting with the OpenAI backend) once the approach looks right to you.

hugohe3:
Thanks for the well-scoped writeup, and for confirming the `/images/edits` path actually applies the edit (not a disguised fresh generation) — that verification is exactly what makes this easy to reason about.

[...]

PR #233, opened 2026-07-13
feat(image-gen): add --reference-image edit support (OpenAI backend)

WodenJay:
Hey, long time no see.

This lands the v1 you scoped in #222: image-to-image editing as an **opt-in, single-image-only** enhancement, OpenAI backend only. No flag → today's behavior is byte-for-byte unchanged, and the `--manifest` / `image_prompts.json` / §VIII / `--render-md` contract is completely untouched — the pipeline-integration question stays out of scope, exactly as you asked.

One honest caveat up front: **I didn't test this by actually editing an image, cuz I don't have an OpenAI API key.** I believe that this PR doesn't have bug, but a real test is needed. Could you plz help me test a real edit? Thanks.

## What changed

`skills/ppt-master/scripts/image_backends/backend_openai.py`
- `SUPPORTS_REFERENCE_IMAGE = True` marker so the dispatcher knows this backend can edit.
- `_image_edits_url()` — resolves the edits endpoint from three base forms: `/v1` (append), already `/images/edits` (as-is), and `/images/generations` (swap the sibling, no double-append).
- `_post_image_edit()` — multipart POST via `files=` so `requests` builds the `Content-Type` boundary itself; the file is opened in a context manager. Uses the `image[]` field for GPT image models and singular `image` for dall-e-2, with a detected MIME type.
- `_edit_image()` — a path parallel to `_generate_image()`, reusing the existing size/quality/format helpers (including `OPENAI_MODERATION`). The existing generation request/response path is left unchanged.
- `generate()` gains an optional `reference_image`; when set, the retry loop calls `_edit_image()` instead of `_generate_image()`.

`skills/ppt-master/scripts/image_gen.py`
- New `--reference-image PATH`, single-image mode only. When set, the prompt is used as the edit instruction.
- Rejects (clean `parser.error`) combining it with `--manifest`, `--render-md`, or `--list-backends`; a missing prompt; or a missing file.
- Rejects a backend without `SUPPORTS_REFERENCE_IMAGE` with a clear message and non-zero exit — no silent fallback to plain text-to-image.
- Positional `prompt` default moved to `None` and backfilled to `"a beautiful landscape"` only in the plain generation branch, so no-flag behavior is identical while edit mode still requires an explicit instruction.

Deliberately **not** in this PR (follow-ups if/when the primitive proves out): manifest / Strategist §VIII integration, masks / inpainting, multi-reference, `input_fidelity`, URL reference input, other backends.

## Why

`image_gen.py` and every backend only did text-to-image; there was no way to pass a source image for editing. This adds the primitive for two manual use cases from the issue — iterate on a previously generated image, and restyle a user-provided source — without touching the automated flow.

## Usage

```bash
python3 skills/ppt-master/scripts/image_gen.py \
  "make the background deep navy and add subtle grain" \
  --reference-image path/to/source.png \
  --backend openai -o out/
```

## How I verified

Local, no network, no real key (monkeypatched `requests.post`), plus subprocess CLI checks — 27 checks, all passing:
- `_image_edits_url()` for all three base forms; `SUPPORTS_REFERENCE_IMAGE` present.
- Edit path hits `/images/edits`, sends multipart with the reference bytes actually uploaded (via `image[]` for GPT image models, singular `image` for dall-e-2, with a MIME type), no JSON body, correct `prompt`/`model`/`size`, and forwards `moderation` from `OPENAI_MODERATION`.
- Generation path still posts JSON to `/images/generations` with the same request shape (regression guard).
- CLI rejects: reference + `--list-backends` / `--manifest` / `--render-md`, missing prompt, missing file, and an unsupported backend.
- `--help` lists `--reference-image`.
- `py_compile` clean on both files; `git diff --check` clean.

**Not verified:** a real end-to-end edit against a live OpenAI/proxy endpoint (no key on my side — help wanted, see above).

## Confirmations (per CONTRIBUTING.md)

- [x] I read `CONTRIBUTING.md` in full before opening this PR.
- [x] I read the full diff myself and ran the affected script(s) locally.
- [x] I confirmed the problem described actually exists in this repository (text-to-image only; no reference-image path on `image_gen.py` / `backend_openai.py`).

Refs #222

### DIFF
--- a/skills/ppt-master/scripts/image_backends/backend_openai.py
+++ b/skills/ppt-master/scripts/image_backends/backend_openai.py
@@ -17,6 +17,11 @@ Configuration keys:
   OPENAI_BACKGROUND          (optional) auto or opaque for gpt-image-2
   OPENAI_MODERATION          (optional) auto or low for GPT image models
 
+Image editing (image-to-image):
+  When image_gen.py passes reference_image=<path> (single-image CLI only,
+  via --reference-image), this backend calls /v1/images/edits with the source
+  image + the prompt as an edit instruction, instead of /v1/images/generations.
+
 Dependencies:
   pip install requests Pillow
 """
@@ -38,6 +43,7 @@ if __name__ == "__main__":
     raise SystemExit(0 if any(arg in {"-h", "--help", "help"} for arg in sys.argv[1:]) else 1)
 
 import base64
+import mimetypes
 import os
 import time
 import threading
@@ -170,6 +176,10 @@ OPENAI_QUALITY_VALUES = {
 GPT_IMAGE_BACKGROUNDS = {"auto", "opaque", "transparent"}
 GPT_IMAGE_MODERATION_VALUES = {"auto", "low"}
 DEFAULT_BASE_URL = "https://api.openai.com/v1"
+
+# Signals to image_gen.py that this backend can accept a reference_image
+# (image-to-image edit via /v1/images/edits). Other backends omit this marker.
+SUPPORTS_REFERENCE_IMAGE = True
 
 
 def _field(value, name: str):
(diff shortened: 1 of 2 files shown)


hugohe3:
Thanks for the contribution — merged!

I pushed a small follow-up in 1ee37e4a to keep image-edit requests model-aware:

- DALL·E 2 edits now omit the GPT Image-only `quality` field.
- GPT Image quality handling and the existing generation path remain unchanged.
- The multipart split stays `image` for DALL·E 2 and `image[]` for GPT Image models.

Validation completed:

- Syntax, CLI, and request-construction smoke checks pass for both `dall-e-2` and `gpt-image-2`.
- Live calls against the configured OpenAI-compatible upstream were inconclusive: `gpt-image-2` timed out after 300 seconds, while `dall-e-2` returned an upstream 503. The local request path is covered; a successful provider-side run is still needed to confirm end-to-end output in that environment.

Thanks again for adding `--reference-image` support.

PR #255, opened 2026-08-10
feat(executor): five-page lightweight spec_lock re-read for Default Generate

Felictycf:
## Summary

Per maintainer scope agreed in #254: implement the **five-page lightweight Spec Lock re-read** for **Default Generate only**.

## Changes (2 files, prompt-only)

1. **`references/executor-base.md` §2.1** — added the scheduled exception to the owning "valid context does not re-read" rule, plus the new rule:
   - After completing P05, P10, P15, … and only when another page follows, read `spec_lock.md` in full once before starting the next page
   - Pure context re-anchor (palette / typography / icon style / page_rhythm)
   - No checker run, no per-page self-check output, no pause, no mid-deck repair loop
   - Compaction / resume / restart still use the full recovery reads
   - Quick Generate has no project root `spec_lock.md` and does not use this rule

2. **`workflows/generate-pptx.md`** — one-line sync pointing to §2.1.

## Scope exclusions (as agreed)

- ❌ No per-page self-check
- ❌ No P2 icon-coverage checker
- ❌ No Visual Review opt-in change
- ✅ Existing P01 first-page gate and final quality gate unchanged; no new mid-deck gate

## Validation

- `prompt_audit.py`: all route budgets PASS; no new audit findings (the 3 pre-existing baseline errors — corpus budget and two load-set limits — are unchanged from before this PR)
- Diff is strictly additive to the two prompt files

### DIFF
--- a/skills/ppt-master/references/executor-base.md
+++ b/skills/ppt-master/references/executor-base.md
@@ -49,7 +49,7 @@ Before the first SVG page, output a confirmation listing: the compact communicat
 
 ### 2.1 Execution context validity (Mandatory)
 
-- **Valid**: if the exact complete Design Spec and lock remain in the unchanged, uncompacted active context, reuse both for every page. Do not reread or poll them.
+- **Valid**: if the exact complete Design Spec and lock remain in the unchanged, uncompacted active context, reuse both for every page. Do not reread or poll them. The one scheduled exception is the five-page lightweight lock re-read defined below.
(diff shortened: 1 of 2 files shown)


hugohe3 (approved):
Scope matches the agreed boundary in #254. The final merge tree against current main passes prompt_audit with 0 errors and 0 warnings.

hugohe3:
Merged in `45bb595a`. Thank you for the focused follow-up to #254.

A small maintainer follow-up landed in `4ad47a16` to make the approved boundaries explicit:

- skip the scheduled re-read when no page follows;
- route an externally changed lock through the existing invalid-context recovery;
- describe Quick as not reading the project-root lock, rather than assuming the file cannot exist.

The final `prompt_audit.py --json` result is 0 errors and 0 warnings.

Issue #262, opened 2026-08-13
Add docs or a validation helper for custom .pptx templates


yulinlina:
The README mentions support for using your own `.pptx` templates, which is one of the most useful parts of this project, but it is not very clear how a template needs to be structured for reliable results.

I was trying this with a company deck and wasn't sure whether PPT Master expects specific slide layouts, placeholder types, or naming conventions. When a template doesn't line up, it seems easy to end up with content falling back to default layouts or appearing in unexpected places, and right now there isn't much guidance for debugging that.

It would help a lot to have a short template-authoring doc covering things like which layouts/placeholders are used for title, body, charts, tables, and speaker notes. Even a small sample template would make onboarding much easier.

Another possibility is a lightweight CLI check like `python -m ppt_master check-template my_template.pptx` that reports missing or incompatible placeholders before generation. That would save users from trial-and-error with generated decks.

hugohe3:
Thanks for raising this. I think the issue may be combining two different workflows:

1. **Fill Native PPTX**: use an existing `.pptx` as a slide library and replace its content. This does not require specific placeholder names or types; PPT Master analyzes and reuses the existing slide shapes.
2. **Create Template → Generate**: convert a PPTX into a reusable template workspace for generating new presentations.

There is no universal placeholder schema that applies to both workflows, so a generic `check-template` command may be misleading. Also, Fill Native PPTX should not silently fall back to default layouts.

If you observed that behavior, could you provide:

- A sanitized source PPTX
- The exact prompt or command used
- The PPT Master version or commit
- The generated output or screenshots

That will help us determine whether this is a routing/documentation issue or an actual bug.

yulinlina:
You’re right — I was conflating the two workflows. The behavior I saw was in **Create Template → Generate**, not Fill Native, so a generic `check-template` would indeed be misleading. I made a sanitized repro instead of sharing the company deck.

The problematic template has a “Title and Content”-like slide where the body is a styled plain text box rather than a real placeholder. Title is a real title placeholder; notes are present as a normal notes text frame.

```python
# make_sanitized_template.py
from pptx import Presentation
from pptx.util import Inches

prs = Presentation()
slide = prs.slides.add_slide(prs.slide_layouts[1])  # Title and Content

# Remove the default body placeholder so the body is just a text box
for ph in list(slide.placeholders):
    if ph.placeholder_format.idx == 1:
        ph._element.getparent().remove(ph._element)

[...]

PR #265, opened 2026-08-14
Add template authoring and validation docs for custom PPTX

yulinlina:
Adds a dedicated template-authoring page and links it from the English and Chinese docs indexes. It clarifies the Fill Native PPTX and Create Template → Generate routes, recommends layout habits, and points users to the existing read-only intake/manifest reports instead of introducing a misleading generic check-template command. Addresses #262.

Addresses #262

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,7 @@ User-facing documentation lives in this directory: English files are the canonic
 | [Audio Narration](./audio-narration.md) | Speaker notes to per-slide narration: providers, voice cloning, timing, PPTX embedding |
 | [Transitions & Animations](./animations.md) | Default behavior and customization for page transitions and per-element object animations |
 | [Templates Guide](./templates-guide.md) | Creating and applying brand / style / layout / deck templates |
+| [Template Authoring](./template-authoring.md) | Custom `.pptx` template checklist and validation with read-only intake/manifest reports |
 
 ## Architecture & Internals
 
--- a/docs/template-authoring.md
+++ b/docs/template-authoring.md
@@ -1,0 +1,71 @@
+# Template Authoring & Validation
+
+[English](./template-authoring.md) | [中文](./zh/template-authoring.md)
+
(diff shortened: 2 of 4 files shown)


hugohe3:
Thanks for putting this together, but I'm closing it unmerged.

1. The PR template's three confirmation checkboxes are unchecked. Per CONTRIBUTING that closes a PR without detailed review on its own.
2. The content restates the FAQ entry that landed in `3fd9aace` and closed #262 — the same two routes, the same two read-only reports, the same note that a styled plain text box is a fillable slot. A second page carrying that in two languages adds ongoing sync burden without adding information.
3. "That PR we talked about" doesn't match the thread. #262 says the opposite: *"there's nothing here to hand off as a PR... neither would a docs PR restating what the docs already say."*

One concrete signal worth noting: `docs/zh/template-authoring.md` copies the English header link line verbatim, so inside `docs/zh/` both language links point at paths that don't exist. Checking the rendered output locally catches that in seconds.

No hard feelings about the original issue — the question was worth asking, and the doc pointer it surfaced is on `main`. The bar for a PR here is a defect in this repository that you verified by running it.

yulinlina:
You’re right on all three points. The PR template checkboxes were not completed, and marking it ready without satisfying that process requirement was wrong. The added page also duplicates the FAQ content introduced in `3fd9aace` rather than adding a new verified gap, and my reference to #262 misread the conclusion there.

The `docs/zh/template-authoring.md` link issue is a fair signal: I did not validate the pages as rendered docs or check language-specific paths. A docs change in this repo should at least pass a local render/link check, especially when separate `en`/`zh` trees are involved.

I won’t ask to reopen this. The useful takeaway is that a contribution here needs to start from a reproducible defect or missing behavior in the current tree, not from rephrasing existing documentation. Going forward, I’d use a stricter checklist:

1. Reproduce the problem on current `main`.
2. Check whether the FAQ or existing intake/manifest docs already cover it.
3. For docs changes, render locally and verify language-specific links.
4. Complete the PR checklist only when each statement is actually true.

Thanks for the specific feedback; it makes the contribution bar clear.

PR #278, opened 2026-09-01
fix: verify wrapped Python daemon processes

### LINKED ISSUE #277: Windows daemon startup fails when launcher PID differs from server PID
### Problem area

Other

### Symptom

On Windows, when PPT Master is run from a virtual environment whose `python.exe` is a launcher/wrapper, both user-facing daemon launchers can report a startup failure even though the Flask server has started and its health endpoint is responding correctly:

- `skills/ppt-master/scripts/confirm_ui/server.py::_wait_for_server_ready`
- `skills/ppt-master/scripts/svg_editor/server.py::_wait_for_ready`

With a uv-created virtual environment, the commands time out with errors such as:

```text
live preview did not become ready at http://127.0.0.1:56123/api/health within 15s (last error: health response belongs to another service or project)
live preview failed to become reachable: http://127.0.0.1:56123
```

Confirm UI fails in the same way after its 10-second startup timeout.

During that wait, `/api/health` returns HTTP 200 with the correct `service` and absolute `project` values. The service also writes a live project lock containing its real PID and requested port. The only failed readiness condition is the comparison between the health PID and the launcher's `Popen.pid`.

### Root cause

On Windows, the Python executable in a uv-created virtual environment is a trampoline. The daemon launcher starts that executable with `subprocess.Popen`, but the trampoline runs the real Python interpreter as a child process.

As a result:

```text
launcher Popen.pid != Flask process os.getpid()
```

The current readiness checks require strict equality:

```python
data.get('pid') == proc.pid
```

The health endpoint and project lock are written by the real Flask process, so they contain the real server PID rather than the wrapper PID. The strict comparison therefore rejects a healthy server and eventually terminates the startup attempt.

This assumption exists in both Confirm UI and SVG live preview, so fixing only one path would leave the sibling daemon broken.

### Why this should be fixed

This is a generic daemon identity-checking bug, not a request to make uv an official dependency or installation requirement.

The official installation path can remain `pip install -r requirements.txt`. No uv package, lockfile, installer change, README recommendation, or uv-specific runtime branch is needed.

The maintainer has already clarified in [Issue #111](https://github.com/hugohe3/ppt-master/issues/111#issuecomment-4471010727) that users who need `venv`, Conda, or uv can run `pip install -r requirements.txt` and subsequent Python commands inside that environment, and that the project does not restrict that path or pin execution to one interpreter. The [roadmap](https://github.com/hugohe3/ppt-master/blob/main/docs/roadmap.md#uv-as-default--required-dependency) separately states that uv will not become the default or a required dependency. The [Windows installation guide](https://github.com/hugohe3/ppt-master/blob/main/docs/windows-installation.md) also notes that Anaconda or Miniconda Python works.

Therefore, a valid alternative Python environment can expose this incorrect PID assumption even though PPT Master does not adopt that environment manager as part of its official toolchain.

The existing project lock protocol already provides a narrow, tool-agnostic identity check: readiness can require the correct health `service` and `project`, the health PID to match the lock PID, the lock port to match the requested port, and that PID to be alive. The existing `proc.poll()` check can remain to detect a launcher that exits abnormally before readiness.

### Steps to reproduce

. On Windows, clone the current repository and create a uv virtual environment:

   ```powershell
   git clone https://github.com/hugohe3/ppt-master.git
   cd ppt-master
   uv venv
   uv pip install -r requirements.txt
   ```

2. Create a minimal live-preview project:

   ```powershell
   $project = Join-Path $env:TEMP "ppt-master-uv-daemon-repro"
   Remove-Item -Recurse -Force $project -ErrorAction SilentlyContinue
   New-Item -ItemType Directory -Force "$project\svg_output" | Out-Null
   ```

3. Start SVG live preview as a daemon:

   ```powershell
   .\.venv\Scripts\python.exe skills\ppt-master\scripts\svg_editor\server.py $project --live --daemon --port 56123 --no-browser
   ```

4. Observe that the command waits for approximately 15 seconds and exits non-zero with `health response belongs to another service or project`, even though the server started and created:

   ```text
   %TEMP%\ppt-master-uv-daemon-repro\live_preview\lock.json
   ```

5. The same failure occurs for Confirm UI when `confirm_ui/server.py <project> --daemon` is run during a normal Default generation after valid Stage-1 inputs have been prepared.

Environment used for this reproduction:

```text
uv 0.12.5
CPython 3.13.13
Windows 11 build 22631.4317
Upstream main: 4e6fdc50136c3aea64a746f6bb4adf1c3305ec87
```

### Expected behavior

Both daemon commands should return success when the correct project service is healthy, owns the matching project lock and requested port, and its recorded server PID is alive, even if a legitimate Python launcher/wrapper has a different PID.

Readiness must still reject a wrong service, wrong project, mismatched or stale lock PID, or mismatched lock port. Shutdown should continue to stop the real server process recorded in the lock.

### AI editor / tool

pi, but this is unrelevant

### AI model

gpt-5.6-sol, but this is unrelevant

### OS

Windows 11, build 22631.4317

### Python version

Python 3.13.13 (uv-created virtual environment; uv 0.12.5)

### Screenshots or files

_No response_

WodenJay:
## What & why

On Windows, a Python executable may be a launcher or wrapper rather than the process that ultimately runs the Flask server. A uv-created virtual environment provides a deterministic reproduction: its `Scripts\python.exe` trampoline starts the real interpreter as a child process.

Both daemon readiness checks assumed that the launcher PID returned by `subprocess.Popen` must equal the PID reported by the Flask process:

```text
launcher Popen.pid == health os.getpid()
```

That assumption fails for wrapped Python executables. The server starts normally, returns HTTP 200 with the correct service and project, and writes its real PID and port to the existing project lock, but the launcher rejects it after the startup timeout because the two PIDs differ.

The affected paths were:

- `confirm_ui/server.py::_wait_for_server_ready`
- `svg_editor/server.py::_wait_for_ready`

This change binds readiness to the existing daemon identity protocol instead. Each launcher now requires:

- an HTTP 200 health response;
- the expected service and absolute project path;
- a health PID matching the PID in the project lock;
- a lock port matching the requested port; and
- a live server PID.

The existing `proc.poll()` check remains in place so a launcher that exits abnormally before readiness is still detected.

This is a focused daemon bug fix, not an adoption of uv. It adds no dependency, lockfile, installer change, README recommendation, or uv-specific runtime branch. The official `pip install -r requirements.txt` path is unchanged; uv is only the reproduction environment for a generic wrapped-executable PID assumption.

Closes #277

## Verification

Environment:

```text
Windows 11 build 22631.4317
uv 0.12.5
CPython 3.13.13
```

Before the change, both real daemon commands reproduced the failure:

- Confirm UI exited non-zero after 10 seconds with `health response belongs to another service or project`.
- SVG live preview exited non-zero after 15 seconds with the same error.

After the change:

- Confirm UI started successfully under the uv virtual environment, returned matching health/lock identity, and `--shutdown` stopped the real server process. The observed launcher PID was `49800`; the health/lock server PID was `63548`.
- SVG live preview started successfully with `--live --daemon`, returned matching health/lock identity, and `--shutdown` stopped the real server process. The observed launcher PID was `11216`; the health/lock server PID was `61584`.
- Both daemons also started and shut down successfully with the non-trampoline base Python interpreter.
- A temporary protocol smoke harness confirmed that both readiness checks still reject a wrong service, wrong project, mismatched lock port, and stale lock PID. The harness was deleted after the run; no test files or test framework were added.

Commands/checks completed:

```text
python -m py_compile skills/ppt-master/scripts/confirm_ui/server.py skills/ppt-master/scripts/svg_editor/server.py
python -m compileall -q skills/ppt-master/scripts
python skills/ppt-master/scripts/attribution_guard.py
PYTHONUTF8=1 python -m unittest discover -s skills/ppt-master/scripts/tests -p "test_*.py"
```

Result: the existing test suite passed, 40/40.

## Confirmations

- [x] This PR is code-only, **or** it touches prompt/instruction text (`SKILL.md`, `references/*.md`, `workflows/*.md`, …) and was discussed and agreed in an issue first — link it here: #277

### DIFF
--- a/skills/ppt-master/scripts/confirm_ui/server.py
+++ b/skills/ppt-master/scripts/confirm_ui/server.py
@@ -1062,12 +1062,17 @@ def _wait_for_server_ready(
         try:
             with urllib.request.urlopen(health_url, timeout=1) as resp:
                 data = json.load(resp)
+                lock = _read_lock(project_path / LOCK_FILE_NAME)
+                server_pid = _lock_pid(lock)
                 if (
                     resp.status == 200
                     and isinstance(data, dict)
                     and data.get('service') == 'confirm_ui'
                     and data.get('project') == str(project_path)
-                    and data.get('pid') == proc.pid
+                    and lock is not None
+                    and lock.get('port') == port
+                    and data.get('pid') == server_pid
+                    and _process_alive(server_pid)
                 ):
                     return True
                 last_error = 'health response belongs to another service or project'
--- a/skills/ppt-master/scripts/svg_editor/server.py
+++ b/skills/ppt-master/scripts/svg_editor/server.py
@@ -1077,12 +1077,17 @@ def _wait_for_ready(
         try:
             with urllib.request.urlopen(health_url, timeout=1) as response:
                 data = json.load(response)
+                lock = _read_lock(_lock_file(project_path))
+                server_pid = _lock_pid(lock)
                 if (
                     response.status == 200
                     and isinstance(data, dict)
                     and data.get('service') == 'live_preview'
                     and data.get('project') == str(project_path)
-                    and data.get('pid') == proc.pid
+                    and lock is not None
+                    and lock.get('port') == port
+                    and data.get('pid') == server_pid
+                    and _process_alive(server_pid)
                 ):
                     return True
                 last_error = 'health response belongs to another service or project'


hugohe3:
Thanks for this — a clean diagnosis and a well-scoped fix. Merged as-is.

I pushed a small follow-up on top (8c9d725e): the readiness waiters now return the lock pid, so the `started ... (pid=...)` line names the real server process instead of the launcher (it would have disagreed with the lock and `--shutdown` in the trampoline case), and the extra `_process_alive` check is dropped since the server has just answered the health probe. Behaviour is otherwise unchanged from your patch.

PR #282, opened 2026-09-05
fix(scripts): keep source converter help dependency-free

### LINKED ISSUE #281: Source conversion backends cannot show --help without optional dependencies
## Problem area

Source conversion (PDF / PPTX / URL to Markdown)

## What happened?

The directly runnable URL, PDF, and PowerPoint conversion backends import their third-party dependencies before `argparse` handles `--help`. In an environment where those optional packages have not been installed yet, users cannot inspect the commands that would tell them how to use the converter:

- `web_to_md.py --help` exits 1 with the `requests` / `beautifulsoup4` dependency message.
- `pdf_to_md.py --help` exits 1 with the PyMuPDF dependency message.
- `ppt_to_md.py --help` exits 1 with a `ModuleNotFoundError` traceback for `pptx`.

This also conflicts with the repository CLI convention that help is handled before optional dependency imports or other side effects.

## Steps to reproduce

From the repository root at `a71a305f`, use `-S` to simulate a standard-library-only Python environment:

```powershell
python -S skills/ppt-master/scripts/source_to_md/web_to_md.py --help
python -S skills/ppt-master/scripts/source_to_md/pdf_to_md.py --help
python -S skills/ppt-master/scripts/source_to_md/ppt_to_md.py --help
```

## Expected behavior

Each command should print its complete argparse help and exit 0 without importing optional conversion dependencies. A real conversion attempt without its required package should still exit non-zero with concise installation guidance and no traceback.

## Suggested fix

Load backend dependencies only after argument parsing, while retaining guarded loading for direct conversion API calls.

## Environment

- AI editor / tool: Codex desktop
- OS: Windows 11
- Python: 3.14.7

QiuLsG:
Fixes #281

## What & why

The standalone web, PDF, and PowerPoint conversion backends imported optional third-party packages before argument parsing. In an environment without those packages, `-h` and `--help` exited during module import instead of displaying usage information.

This change:

- loads conversion dependencies only after CLI arguments have been validated;
- keeps all three backend modules importable without optional site packages;
- reports concise installation guidance, without a traceback, when a real conversion lacks its dependency; and
- preserves web, PDF, and PowerPoint conversion behavior when dependencies are installed.

Reproduction before this change:

```text
python -S skills/ppt-master/scripts/source_to_md/web_to_md.py --help
python -S skills/ppt-master/scripts/source_to_md/pdf_to_md.py --help
python -S skills/ppt-master/scripts/source_to_md/ppt_to_md.py --help
```

Each command stopped at an eager third-party import rather than reaching `argparse` help. With this change, both `-h` and `--help` work for every backend without site packages.

## Verification

- Ran `-h` and `--help` under `python -S` for all three backends: all six calls exited 0 with clean help output.
- Exercised dependency-missing conversion paths: each exited 1 with concise installation guidance and no traceback.
- Imported all three modules without site packages and confirmed third-party modules were not loaded eagerly.
- Converted `https://example.com` with the web backend.
- Converted temporary generated PDF and PPTX fixtures with the PDF and PowerPoint backends.
- Ran `python -m py_compile` for all three changed files.
- Ran `skills/ppt-master/scripts/tests/test_source_to_md_cli.py`: 8 tests passed.
- Ran `git diff --check`.

## Confirmations

### DIFF
--- a/skills/ppt-master/scripts/source_to_md/pdf_to_md.py
+++ b/skills/ppt-master/scripts/source_to_md/pdf_to_md.py
@@ -5,6 +5,8 @@ Uses PyMuPDF to extract PDF text content and convert to Markdown format.
 Supports heading levels, bold, italic, and list detection.
 """
 
+from __future__ import annotations
+
 import argparse
 import hashlib
 import json
@@ -24,11 +26,17 @@ from _conversion_profile import write_conversion_profile_best_effort  # noqa: E4
 
 configure_utf8_stdio()
 
-try:
-    import fitz  # PyMuPDF
-except ImportError:
-    print("[ERROR] PyMuPDF not installed. Run: pip install PyMuPDF", file=sys.stderr)
-    sys.exit(1)
+# Help must not depend on the optional conversion packages: a stdlib-only
+# interpreter still gets the argparse usage (docs/rules/code-style.md §4).
+_HELP_REQUESTED = __name__ == "__main__" and any(
+    arg in {"-h", "--help"} for arg in sys.argv[1:]
+)
+if not _HELP_REQUESTED:
+    try:
+        import fitz  # PyMuPDF
+    except ImportError:
+        print("[ERROR] PyMuPDF not installed. Run: pip install PyMuPDF", file=sys.stderr)
+        sys.exit(1)
 
 FONT_BODY_SIZE = 12
 FONT_H1_SIZE = 24
@@ -1746,7 +1754,7 @@ def extract_pdf_to_markdown(
     return markdown_content
 
 
-def main() -> int:
+def main(argv: list[str] | None = None) -> int:
     """Run the CLI entry point."""
     parser = argparse.ArgumentParser(
         description='PDF to Markdown converter (with structure detection and LLM optimization)',
@@ -1793,7 +1801,7 @@ Structure detection features:
         help=f'DPI for --render-vector-figures output (default: {VECTOR_FIGURE_DPI})',
     )
 
-    args = parser.parse_args()
+    args = parser.parse_args(argv)
 
     return run_path_batch(
         args.inputs,
--- a/skills/ppt-master/scripts/source_to_md/ppt_to_md.py
+++ b/skills/ppt-master/scripts/source_to_md/ppt_to_md.py
@@ -50,12 +50,22 @@ from pptx_ooxml.diagram_read import (  # noqa: E402
     smartart_to_markdown,
 )
 
-from pptx import Presentation
-from pptx.enum.action import PP_ACTION
-from pptx.enum.shapes import MSO_SHAPE_TYPE
-from pptx.oxml.ns import qn
-
 configure_utf8_stdio()
+
+# Help must not depend on the optional conversion packages: a stdlib-only
+# interpreter still gets the argparse usage (docs/rules/code-style.md §4).
+_HELP_REQUESTED = __name__ == "__main__" and any(
+    arg in {"-h", "--help"} for arg in sys.argv[1:]
+)
+if not _HELP_REQUESTED:
+    try:
+        from pptx import Presentation
+        from pptx.enum.action import PP_ACTION
+        from pptx.enum.shapes import MSO_SHAPE_TYPE
+        from pptx.oxml.ns import qn
+    except ImportError:
+        print("[ERROR] python-pptx not installed. Run: pip install python-pptx", file=sys.stderr)
+        sys.exit(1)
 
 
 EMU_PER_INCH = 914400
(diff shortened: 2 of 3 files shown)


hugohe3:
Thanks for the report and the PR. #281 is real: on current `main`, `python3 -S skills/ppt-master/scripts/source_to_md/{web,pdf,ppt}_to_md.py --help` all exit 1 before argparse runs, which breaks the rule in `docs/rules/code-style.md` §4 (help must be handled before optional dependency imports). This is worth fixing, but the PR needs three changes before it can be merged.

**1. It conflicts with `main`.** The branch is based on `a71a305f`. Since then `web_to_md.py` changed on `main` (`fetch_url` now returns the final URL, `traverse` keeps inline spacing and resolves relative links, image handling applies EXIF orientation). GitHub already marks the PR as conflicting, so please rebase onto the current `main` first.

**2. The lazy-loading approach breaks the module as a library.** On `main`, `tests/test_source_to_md_cli.py` imports `web_to_md` and calls `simple_html_to_markdown_traversal(...)` directly, and calls `process_url(...)` with `_http_get` patched out. With the PR, `NavigableString`, `BeautifulSoup`, `Tag`, and `Comment` stay `None` on those paths because `_require_web_dependencies()` only runs from `main()` and `_http_get()`, so `isinstance(node, NavigableString)` raises `TypeError`. The "8 tests passed" result is correct for your base commit; the current module has 30 cases, and several of them exercise exactly these paths.

**3. The repository convention is an early help guard, not module-wide rebinding.** `docs/rules/code-style.md` §4 asks for `-h`/`--help` to be handled by argparse or an explicit early guard *before* any side effect, and it deliberately keeps the eager `try: import ... except ImportError: <install hint>; sys.exit(1)` block so that library imports and real conversions keep their current semantics. The smallest conformant fix is roughly:

```python
configure_utf8_stdio()

if __name__ == "__main__" and any(arg in {"-h", "--help"} for arg in sys.argv[1:]):
    build_parser().print_help()
    raise SystemExit(0)

try:
    import requests
    from bs4 import BeautifulSoup, Comment, NavigableString, Tag
except ImportError:
    ...
```

with the `argparse.ArgumentParser(...)` construction moved out of `main()` into a `build_parser()` function so the guard and `main()` share it. The same shape applies to `pdf_to_md.py` and `ppt_to_md.py`. That keeps the diff small, leaves `fitz` / `pptx` / `bs4` as ordinary module globals, and still gives a concise install message (no traceback) when a real conversion runs without its dependency.

If you prefer to keep the lazy loaders, the alternative is to call `_require_web_dependencies()` at the top of every public function that touches bs4 (`simple_html_to_markdown_traversal`, `process_url`, `element_to_markdown`, ...), but the guard above is what the rest of the scripts do.

**How to resubmit**

1. Rebase onto `main` and resolve `web_to_md.py` against the current version.
2. Apply the early help guard (or the full set of `_require_*` calls) to all three backends.
3. Verify from the repository root:

   ```bash
   python3 -S skills/ppt-master/scripts/source_to_md/web_to_md.py --help
   python3 -S skills/ppt-master/scripts/source_to_md/pdf_to_md.py --help
   python3 -S skills/ppt-master/scripts/source_to_md/ppt_to_md.py --help
   cd skills/ppt-master/scripts && python3 -m unittest tests.test_source_to_md_cli
   ```

   All three help calls should exit 0 and the test module should report 30 passing cases.
4. Force-push the branch; the PR updates in place.

Happy to review again once it is rebased.

hugohe3:
Rebased onto `main` and reworked the fix in place so this can merge with you as the author — thanks for the report and the reproduction.

What changed versus your version:

- Kept the eager dependency imports (library imports and real conversions behave exactly as before) and added one `_HELP_REQUESTED` flag right after `configure_utf8_stdio()`; the optional-dependency `try` block is skipped only when `-h`/`--help` is on the command line, and argparse then prints help and exits 0. This is the early-guard shape from `docs/rules/code-style.md` §4 and keeps the current `test_source_to_md_cli` cases (which import `web_to_md` directly) working.
- `ppt_to_md.py` gains the concise `python-pptx` install hint you proposed instead of a traceback.
- `pdf_to_md.py` / `ppt_to_md.py` `main()` now take `argv`, and the `urllib3.disable_warnings` call in `web_to_md.py` is skipped on the help path.
- `from __future__ import annotations` added to `web_to_md.py` / `pdf_to_md.py` as in your PR, so `fitz` / `Tag` annotations do not evaluate at import time.

Verified with `python3 -S` for all three backends (`-h` and `--help` exit 0), missing-dependency runs exit 1 with the install hint, and the 30 `test_source_to_md_cli` cases pass.

PR #283, opened 2026-09-07
fix(video-subtitles): keep the space between merged subtitle clauses

MaxFreedomPollard:
## What & why

`_split_sentence()` in `skills/ppt-master/scripts/video_subtitles.py` cuts an overlong narration sentence at clause punctuation, strips each clause, and then merges short neighbours back together while they still fit inside `--max-chars`. The merge on `main` is `candidate = f"{output[-1]}{atom}"` (line 175), and by that point the space that separated the two clauses has already been stripped off, so a cue that merges `We plan,` with `we build,` is published as `We plan,we build,`.

I ran into this on an English deck. Chinese narration never shows it, because there is no space between clauses to lose, which is probably why it went unnoticed.

The guard at the end of `align_video_subtitles()` cannot catch it either: it compares `_text_key()` of the aligned output against `_text_key()` of the frozen transcript, and `_text_key()` removes all whitespace, so the two sides compare equal and the SRT is published.

The same algorithm already lives in `skills/ppt-master/scripts/tts_backends/backend_edge.py` as `_split_sentence_span()` and `_hard_split_span()`, and that copy is correct: it carries `(start, end)` spans and merges with `candidate = (merged[-1][0], atom[1])`, so every cue stays a verbatim slice of the source text and whatever separated the clauses survives. This PR gives `video_subtitles.py` the same shape. Clause splitting, hard splitting, and merging now work on spans of the normalized sentence, and the text is sliced only when the lines are returned. Split points and cue counts are untouched, only the spacing inside a merged cue changes.

Reproduction from the repository root:

```
$ python3 -c "import sys; sys.path.insert(0, 'skills/ppt-master/scripts'); import video_subtitles as v; print(v._split_sentence('We plan, we build, we ship.', 20))"
['We plan,we build,', 'we ship.']      # main
['We plan, we build,', 'we ship.']     # this branch
```

The same sentence through `backend_edge._subtitle_cues()` gives `['We plan, we build,', 'we ship.']` on `main`, which is what convinced me the string-concatenating copy is the one that is wrong.

## Verification

Real-project run, on macOS 26.2 with Python 3.11.15 and ffmpeg 8.1.1. I built a project with two page-local SRT files (one English, one Chinese) under `audio/`, made a six second video with `ffmpeg -f lavfi -i color=c=navy:s=640x360:d=6`, and ran the CLI end to end:

```
python3 skills/ppt-master/scripts/video_subtitles.py <project> --video exports/demo.mp4 --language en --force
```

Both runs report `Aligned 7 final-video subtitle cue(s); last cue ends at 14.000s`. `diff` between the two published `exports/demo.srt` files is one line: cue 1 is `We plan,we build,` on `main` and `We plan, we build,` on this branch. Cues 2 to 7 are identical, including the two Chinese cues, and so are all the timings. stable-ts is not installed here, so `stable_whisper` was a local stub on `PYTHONPATH` that echoes the transcript back as an SRT, one cue per line, which is what `original_split=True` asks the real library to do. Everything the change touches runs before that call, and the whole publish path after it (the `_text_key()` comparison, the cue-count check, the temp file and `os.replace`) ran for real.

Differential check against the old code, to make sure I only changed spacing: I kept a verbatim copy of the `main` version of `_hard_split()` and `_split_sentence()` and compared it with the new one over 59428 random strings built from Latin letters, Chinese characters, spaces, tabs, and the six clause-ending marks, with `max_chars` from 1 to 12. Zero differences in cue count, zero differences in the whitespace-stripped text, and 18411 inputs where a dropped space came back. I also asserted on every output line that it is a substring of the whitespace-normalized sentence and carries no leading or trailing whitespace.

Tests, following `docs/rules/code-style.md` §11:

```
cd skills/ppt-master/scripts && python3 -m unittest tests.test_video_subtitles
```

5 tests, OK. Against `main` the same module fails 4 of the 5, and the CJK case passes, which is the split I wanted. I also ran all seventeen modules under `scripts/tests/` one at a time. Sixteen pass. `tests.test_source_to_md_cli` fails 17 of its 23 tests, identically on `main` and on this branch, and it is unrelated to this change: on macOS `tempfile.mkdtemp()` hands back `/var/folders/...` while the code resolves it to `/private/var/folders/...`, so the path assertions compare a resolved path against an unresolved one. Worth a look separately if you develop on macOS.

Also `python3 -m py_compile` on both files, `git diff --check`, and `ruff check` on both files (clean, and the repository ships no lint config).

## Confirmations

- [x] I personally reviewed the full diff and verified every claim in this description against this repository's actual code, including that the problem is real here and the capability isn't already provided by an existing path or already fixed on `main` (purely AI-generated PRs without human review are closed unmerged)
- [x] This PR is code-only, **or** it touches prompt/instruction text (`SKILL.md`, `references/*.md`, `workflows/*.md`, …) and was discussed and agreed in an issue first, link it here: #___

### DIFF
--- a/skills/ppt-master/scripts/tests/test_video_subtitles.py
+++ b/skills/ppt-master/scripts/tests/test_video_subtitles.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env python3
+"""Unit tests for video_subtitles.py cue splitting and clause merging."""
+
+from __future__ import annotations
+
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+SCRIPTS_DIR = Path(__file__).resolve().parents[1]
+if str(SCRIPTS_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPTS_DIR))
+
+import video_subtitles  # noqa: E402
+
+
+PAGE_SRT = """1
+00:00:00,000 --> 00:00:04,000
+We plan, we build, we ship.
+
+2
+00:00:04,000 --> 00:00:08,000
+Design first, then measure, then iterate.
+"""
+
+
+class ClauseMergeTests(unittest.TestCase):
+    def test_merged_clauses_keep_the_space_that_separated_them(self) -> None:
+        self.assertEqual(
+            video_subtitles._split_sentence("We plan, we build, we ship.", 20),
+            ["We plan, we build,", "we ship."],
+        )
+        self.assertEqual(
+            video_subtitles._split_sentence(
+                "Alpha, beta, gamma, delta and epsilon.", 20
+            ),
+            ["Alpha, beta, gamma,", "delta and epsilon."],
+        )
+
+    def test_cjk_clauses_merge_without_inventing_a_space(self) -> None:
+        self.assertEqual(
+            video_subtitles._split_sentence(
+                "第一步是调研，第二步是设计，第三步是交付。", 20
+            ),
+            ["第一步是调研，第二步是设计，", "第三步是交付。"],
+        )
+
+    def test_short_sentence_is_returned_unsplit(self) -> None:
+        self.assertEqual(
+            video_subtitles._split_sentence("We ship, we learn.", 20),
+            ["We ship, we learn."],
+        )
+
+    def test_every_cue_is_a_trimmed_slice_within_the_character_budget(self) -> None:
+        sentences = (
+            "We plan, we build, we ship.",
+            "Design first, then measure, then iterate on the result.",
+            "Alpha, beta, gamma, delta and epsilon.",
+            "第一步是调研，第二步是设计，第三步是交付。",
+            "One extraordinarily long unbreakable token: supercalifragilistic.",
+        )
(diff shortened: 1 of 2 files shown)


hugohe3:
Thanks for this, @MaxFreedomPollard — a clean catch, and the write-up made it easy to verify. I reproduced the `We plan,we build,` cue on `main`, and an independent differential fuzz against your branch (60k random Latin/CJK/whitespace/punctuation strings, `max_chars` 1–12) showed zero changes in cue count or stripped text, only restored spaces, which matches your own numbers.

Merged as 44d32733. One heads-up: I amended the commit message on your branch before merging (author unchanged, code untouched) to tighten two sentences — the CJK note (a Chinese sentence with a space between clauses hits the same bug) and the `backend_edge.py` comparison (it shares the span-and-merge shape but splits on word timings, so it is not the same hard-split algorithm). The diff itself is exactly what you submitted.

Thanks also for flagging the `test_source_to_md_cli` path-resolution failures on macOS; that is noted separately.

PR #287, opened 2026-09-12
fix: open preview URLs in Windows browser under WSL

zhuoyi780-alt:
## What & why

On WSL2 with `gio` on PATH but no default HTTP handler, both preview entry points use Python's Linux `webbrowser` dispatch, so they can serve a page without opening the Windows browser. The SVG editor's native Windows branch does not apply to WSL (`os.name == "posix"`).

Use a shared browser helper for Confirm UI and the SVG editor. On WSL with PowerShell available, request the Windows default browser with `Start-Process`. Encode the URL as data, bound the subprocess wait, and report launch failures with the manual URL. Keep native Windows dispatch and the generic webbrowser path for other hosts or WSL without PowerShell.

Reproduction: on WSL with PowerShell on PATH, `webbrowser.get().name` returns `gio` and `gio mime x-scheme-handler/http` reports no default application. Start either configured preview entry point with `--daemon` and observe that the service can be ready while automatic browser opening fails. Local-network sandbox permissions must already allow the preview service; this patch does not change those permissions.

## Verification

- Rebased on upstream main `f76d1ecb4185a61e107702d6244b41f487834cc4`; all 15 regression tests rerun successfully. Generic smoke tests were completed before the rebase; intervening upstream changes do not touch either preview entry point or their shared helper.
- Python 3.13, Ubuntu on WSL2: 7 browser-dispatch regression tests, 2 existing SVG editor tests, and 6 existing local-server security tests passed (standard-library unittest).
- Ran patched SVG editor with a synthetic SVG containing only generic browser-test text: health, homepage and slide-list endpoints returned HTTP 200. Windows accepted the browser launch request.
- Ran patched Confirm UI with independently generated generic template options and recommendations: health and homepage returned HTTP 200; Windows accepted the browser launch request.
- Both smoke-test services were shut down after verification. The reported smoke tests use synthetic fixtures only.
- Windows and non-WSL Linux dispatch paths were mocked in regression tests; no native Windows/Linux runtime verification is claimed. OS acceptance does not prove window visibility.
- `git diff --cached --check` passed.

## Confirmations

- [x] This PR is code-only; it does not modify prompt/instruction text.

### DIFF
--- a/skills/ppt-master/scripts/confirm_ui/server.py
+++ b/skills/ppt-master/scripts/confirm_ui/server.py
@@ -46,7 +46,6 @@ import time
 import urllib.error
 import urllib.parse
 import urllib.request
-import webbrowser
 from pathlib import Path
 from typing import Optional
 
@@ -68,6 +67,7 @@ from server_common import (  # noqa: E402
     clear_lock as _clear_lock,
     find_free_port as _find_free_port,
     lock_pid as _lock_pid,
+    open_preview_browser,
     popen_detached as _popen_detached,
     process_alive as _process_alive,
     read_lock as _read_lock,
@@ -1134,7 +1134,7 @@ def _launch_background_server(
     url = _server_url(port)
     logger.info('started confirm UI in background: %s (pid=%s)', url, server_pid)
     if open_browser:
-        webbrowser.open(url)
+        open_preview_browser(url, logger=logger)
     return proc, port, log_path
 
 
@@ -1162,7 +1162,7 @@ def _open_browser_async(url: str, delay: float = 0.4) -> None:
     """Open the browser after Flask has had a moment to bind its socket."""
     def _open() -> None:
         time.sleep(delay)
-        webbrowser.open(url)
+        open_preview_browser(url, logger=logger)
 
     threading.Thread(target=_open, daemon=True).start()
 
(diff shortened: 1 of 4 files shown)


hugohe3:
Thanks — the problem is real (CPython's `BackgroundBrowser.open` returns True as soon as `gio open` spawns, so WSL2+WSLg without an HTTP handler fails silently) and the shared-helper shape is the right one. I amended your commit on this branch with a few adjustments so it could merge as-is; you remain the author.

- `BROWSER` is honoured on every host before the WSL branch runs, so users who already configured `wslview` or a WSLg-side browser keep that behaviour.
- `powershell.exe` is looked up on PATH first and then at `/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe`, so `[interop] appendWindowsPath=false` hosts still dispatch instead of falling back to the silent gio path.
- Native Windows goes back to `webbrowser.open`: it already does `os.startfile` and falls back to PATH browsers when the HTTP association is broken, which a bare `os.startfile` loses.
- `open_preview_browser` moved to its alphabetical slot in both import lists.
- Tests now assert the script actually passes the URL to `Start-Process -FilePath`, and cover the `BROWSER` override and the mounted-PowerShell probe.

Verified here: 9 browser tests plus the existing SVG editor / local-server security suites pass; `svg_editor/server.py --daemon` on a real project copy starts, serves `/api/health` 200, logs the manual-URL warning on a display-less Linux host, and shuts down cleanly. Merging.

